PR #1, opened 2022-05-04
Design doc

dtjong:


designs/closures-design.md
```diff
@@ -0,0 +1,143 @@
...
+literal :=
+...
+| lambda [<name> [, <name>]*]? : <expr>
+```
+
+### AST
+
+```ts
+// To stay a subset of Python, lambda parameters cannot have type hints,
+// so we defer the type-checking of an anonymous function to the point of use
```
jpolitz: Oof this seems super tricky to get right. I don't know if this check-at-use behavior is OK. Reporting the error for multiple conflicting uses is tough, and how will type-checking work if the lambda is an argument to an unresolved lambda param?

I wonder what other syntactic hacks are possible. Maybe a special variable-arity function like `mklambda(Callable[...], lambda x, y: ...)` or similar, just to have a place to put the annotation?

dtjong: That's a good point, we don't have a quick fix either and we'll think about it.

jpolitz:
This looks generally solid.

One of the things that I think may need to be coordinated with the memory management group is allocation. They'll need to know which closure-stored variables are references or not. The type of the `Callable` isn't enough to know this, so there needs to be something between the callable type and the generated class that gives this information.

Also, at which point in the compiler pipeline do you expect to insert the generated classes for closures?

Finally, your 10 tests are great but quite expansive. What do you think you'll get done this week (say, for May 13)?

PR #5, opened 2022-05-06
Set deisgn

ShuchiZhang:
Update ast.ts, ir.ts file for set, dict, and tuple. Test cases and the basic idea of implementation for the set is described in design/set-design.md.


.gitignore
```diff
@@ -5,4 +5,6 @@ debugger
 cli
 *.sw*
 mochawesome-report
-*lock*
\ No newline at end of file
+*lock*
+.DS_Store
+.vscode/*
```
jpolitz, line 10: It's weird to have a PR that ignores `.vscode` but also adds a file to it. Did you mean to add `.vscode/launch.json` (do you think everyone should use it?)

README.md
```diff
@@ -0,0 +1,32 @@
+# chocopy-wasm-compiler-A
```
jpolitz, line 1: I don't know that this belongs in the _global_  README for the whole compiler.

ir.ts
```diff
@@ -32,6 +32,10 @@ export type Expr<A> =
   | {  a?: A, tag: "alloc", amount: Value<A> }
   | {  a?: A, tag: "load", start: Value<A>, offset: Value<A> }
 
+  | {  a?: A, tag: "set_expr", contents: Array<Expr<A>> }
```
jpolitz: I really hope that we don't have to add these to `ir`, and can express them with a combination of `call`, `alloc`, `load`, `store`, etc.

tests/types.test.ts
```diff
@@ -1,101 +1,142 @@
```
jpolitz, line 1: It would be useful to move set-specific tests into their own file, like `types.sets.test.ts`. Lots of merge conflicts will result with these files being edited by all.

designs/set-design.md
```diff
@@ -0,0 +1,195 @@
+# Set Dict Tuple Design
+## Change to AST and Ir
+We made changes to ast.ts and ir.ts to include set, dictionary and tuples. Though we will focus on set but we have included all three types in case other group might refer to these types. Changes in ast and ir are the same as we have not included optimization currently.
+
+Similar to class implementation in PA3, we add three new types under the Type we declared as:
+```
+export type Type =
+    ……
+  | { tag: "set"; content_type: Type }
```
jpolitz: If this week is just about sets (which is fine) let's _just_ add the infrastructure for sets.

designs/set-design.md
```diff
@@ -0,0 +1,195 @@
...
+
+Set.add()
+
+Set.remove()
+
+Set.clear()
+
+Set.update()
+
+in keyword (x in Set)
```
jpolitz: How will `in` be parsed and represented?

designs/set-design.md
```diff
@@ -0,0 +1,195 @@
...
+```
+x:set = {1,2,3}
+1 in x
+7 in x
+```
+expected output:
+```
+True
+False
+```
```
jpolitz: What will it look like when a set is given as an argument to a function?

designs/set-design.md
```diff
@@ -0,0 +1,195 @@
...
+TypeError: 'int' object is not iterable
+```
+### 10. in keyword
+Description:
+
+This test case uses 'in' keywords to judge whether an element is in the set.
+
+program:
+```
+x:set = {1,2,3}
```
jpolitz: How will the type-checker know that this is a set of _numbers_? It seems like this annotation needs more information (which you included in `Type` but not in the concrete syntax)

tests/types.test.ts
```diff
@@ -1,101 +1,142 @@
...
+  x.add(1)
+  print(x)`,["{1,2}"])
+});
+   
+describe("proj-dict-test", ()=>{
+    
+  assertTC("tc-assign-dict-int-bool",`
+  x:set = {1:True,2:False}`, DICT(NUM,BOOL))
+
+  assertPrint("pr-dict-int-bool-key",`
```
jpolitz: I would take out tests that you don't expect to pass early on, or somehow make it so they don't run – won't these just fail right now (and you're not planning on making them pass by the end of the week?)

utils.ts
```diff
@@ -34,3 +34,13 @@ export const NUM : Type = {tag: "number"};
 export const BOOL : Type = {tag: "bool"};
 export const NONE : Type = {tag: "none"};
 export function CLASS(name : string) : Type {return {tag: "class", name}};
+// Week6: Project Milestone Tests
+export function TUPLE(content_type : Array<Type>) : Type{
```
jpolitz, line 38: Glad you're thinking about these.

designs/set-design.md
```diff
@@ -0,0 +1,195 @@
...
+Then in the expression, the change we have made is as:
+```
+export type Expr<A> =
+    ……
+ | { a?: A, tag: "set_expr", contents: Array<Expr<A>> }
+ | { a?: A, tag: "tuple_expr", contents: Array<Expr<A>> }
+ | { a?: A, tag: "dict_expr", entries: Array<[Expr<A>, Expr<A>]> }
+```
+## Value Representation and Memory Management
+All Values are stored as 8-byte i32. Memory management should use the interface provided by the memory group, the memory usage should be kept below 30% to maintain the efficiency of the hash function.
```
jpolitz: I feel like this is the only mention of hashing/a hash table. How is that going to work?

jpolitz (commented):
I have a bunch of questions that I left inline in the code/design doc – I like starting with sets, and think there are some useful things to focus on just within those constraints.

PR #6, opened 2022-05-06
Design document (Optimizations)

nirmalthomas2609:
Optimizations
For the first week of the project, we will be implementing constant propagation and constant folding


designs/optimizations-design.md
```diff
@@ -0,0 +1,197 @@
...
+</li>
+
+<li>
+Constant propagation:
+Test case #10
+
+```
+a = 3
+def f(b):
+|  b = a
```
jpolitz, line 171: What is this pipe character?

designs/optimizations-design.md
```diff
@@ -0,0 +1,197 @@
...
+f(3)
+```
+This will not get transformed in the current iteration of constant propagation.
+</li>
+</ul>
+
+## List of changes to AST/IR
+
+We will not be making any changes to the IR/AST. We will need to take in changes to these from all teams.
+Our work will be to just update IR statements values and (or) annotations and also to reduce them.
```
jpolitz, line 186: Note that your examples above are given by describing _Python_ code, but there are no `if`, `while`, etc statements at the IR level. Does that complicate things at all? Might be useful to express these as a few levels of Python -> IR -> IR* (optimized)

jpolitz (commented):
These are good opts to pick to get some infrastructure in place. I'm a little worried that infrastructure isn't described, and the test description I don't know how to turn into something runnable.

What will you do to run these tests in an automated way? That's my biggest question.

PR #8, opened 2022-05-06
List - Project Milestone Tests

AlexfzL:
Test cases we expect to pass next week along with changes that we want to make to some files.


ast.ts
```diff
@@ -26,6 +27,7 @@ export type Stmt<A> =
   | {  a?: A, tag: "field-assign", obj: Expr<A>, field: string, value: Expr<A> }
   | {  a?: A, tag: "if", cond: Expr<A>, thn: Array<Stmt<A>>, els: Array<Stmt<A>> }
   | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
+  | {  a?: A, tag: "item-assign", obj: Expr<A>, idx: number, value: Expr<A> }
```
jpolitz: When/how do you plan to represent statements like

```
x : int = 0
[1, 2, 3][x] = 5
```
?

ast.ts
```diff
@@ -38,6 +40,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "construct-list", items: Array<Literal> } // [1,2,3]
+  | {  a?: A, tag: "get-item", idx: number } // a[0]
```
jpolitz: When/how do you plan to represent statements like

```
x : int = 0
[1, 2, 3][x]
```
?

designs/lists-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+```
+List of lists/objects
+a: [[int]] = None
+b: [[C]] = None
+class C(object):
+    x:int = 3
+    pass
+a = [[0,1,2], [1,2,3]]
+b = [[C(), C()]]
```
jpolitz, line 86: This example doesn't match the AST above – `C()` is not a literal.

designs/lists-design.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+and add to Expr:
+
+```
+{ a?:A, tag: "construct-list", items: Array<Literal> } // [1,2,3]
+{ a?:A, tag: "get-item", idx: number } // a[0]
+```
+
+We also expect to have changes to the globals so we can support a function to print lists, get the length of the list, and also index the element at the correct offset. Since lists are also mutable and expandable, we're thinking of having some notion of the correct offsets or locations in memory that correspond to the list values. We also need to make some changes in parser, typecheck, and compiler that are corresponding to the changes we made in ast and some of the new syntax rules that the data structure list would support.
```
jpolitz, line 143: I think a bunch of detail about the memory representation will be useful. Will lists be resizable in this milestone? How will the runtime representation support list-index-out-of-bounds errors?

It's also important to write some tests where lists are _arguments to functions_, because that forces you to think about lists for which the compiler doesn't know all of the contents statically.

jpolitz (commented):
I want to hear more about the implementation strategy, including memory layout.

Also, I think limiting indices to just number literals is too much of a restriction and will be difficult to manage later on – indices should be able to be expressions (even the simplest for loop over list indices will require this).

I think thinking through these two issues will go a long way towards getting started this week.

Nice work!

lower.ts
```diff
@@ -1,16 +1,15 @@
-import * as AST from './ast';
```
jpolitz: URGENT – this entire file got reformatted so the diff is very large and we can't efficiently review it. Please restore its earlier formatting.

compiler.ts
```diff
@@ -81,6 +80,21 @@ export function compile(ast: Program<Type>, env: GlobalEnv) : CompileResult {
 function codeGenStmt(stmt: Stmt<Type>, env: GlobalEnv): Array<string> {
   switch (stmt.tag) {
     case "store":
+      if(stmt.start.tag === "listaddr") {
```
jpolitz: I think this isn't the right design. I would expect that all of the work for looking up/storing lists can be done in `lower.ts`. You can use `ifjmp` to manage the errors (or add special functions like `assert_not_none`), and use `alloc` and `store` along with intermediate variables to do this work. This will complicate optimization, memory management, and so on to have special type-specific operations within the code generation step.

lower.ts
```diff
@@ -139,6 +139,25 @@ function flattenStmt(s : AST.Stmt<Type>, blocks: Array<IR.BasicBlock<Type>>, env
...
+
+    case "index-assign": {
+      var [oinits, ostmts, oval] = flattenExprToVal(s.obj, env);
+      var [iinits, istmts, ival] = flattenExprToVal(s.index, env);
+      var [ninits, nstmts, nval] = flattenExprToVal(s.value, env);
+
+      if (s.obj.a.tag !== "list") {
+        throw new Error("Compiler's cursed, go home.");
+      }
+      pushStmtsToLastBlock(blocks, ...ostmts, ...istmts, ...nstmts, {
```
jpolitz: Here, for example, I think you could add a few extra statements to check that the index is in bounds and that the list is not none, and then generate a `store` that doesn't require any changes at all to the IR.

lower.ts
```diff
@@ -285,6 +304,69 @@ function flattenExprToExpr(e : AST.Expr<Type>, env : GlobalEnv) : [Array<IR.VarI
...
+    case "index":
+      var [objInit, objStmts, objExpr] = flattenExprToVal(e.obj, env);
+      var [idxInit, idxStmts, idxExpr] = flattenExprToVal(e.index, env);
+      return [
+        [...objInit, ...idxInit],
+        [...objStmts, ...idxStmts],
+        {
+          a: e.a,
+          tag: "load",
+          start: { a: { tag: "number" }, tag: "listaddr", addr: objExpr },
```
jpolitz: Again here, could do bounds checking in lower, and leave IR unchanged.

tests/lists.test.ts
```diff
@@ -0,0 +1,551 @@
+import * as mocha from "mocha";
```
jpolitz, line 1: One of the things we talked about in class is making sure there are some tests that are inside functions – can you please add some tests that have lists as function arguments, or create lists inside of functions?

You actually have a lot of great tests, but these would help make sure a number of things regarding scope and list creation work well.

jpolitz (changes requested):
I'd like to see as much as possible move out of `compiler.ts` and into `lower.ts` to keep the IR compilation simple (and not complicate other low-level work). I'd be happy to hear why that can't be done if I'm misunderstanding.

The other comments are small code comments and additional test suggestions.

Nice work!

lower.ts
AlexfzL: We moved checking index in bound to lower from compiler

Hackerry: Code was fixed to reflect this comment and the next.
We've moved all low-level code generation to lower.ts.

tests/lists.test.ts
Hackerry: We've added new test cases where
1) lists are used as function parameters
2) lists are returned from functions

parser.ts
```diff
@@ -232,38 +242,112 @@ export function traverseExprHelper(c: TreeCursor, s: string, env: ParserEnv): Ex
...
+        if (s.substring(c.from, c.to) == "]") {
+          c.parent();
+          return { tag: "index", obj: objExpr, index: startIndex };
+        }
+
+        // Start index and :
+        c.nextSibling(); // Focus on end index or ]
+        if (s.substring(c.from, c.to) == "]") {
+          c.parent();
+          return { tag: "slice", obj: objExpr, index_s: startIndex };
```
yiz136, line 292: We are set/dict group. We'll probably use your "slice" expr in AST for our dictionary add (Because it's also a "[]" expression and we don't want to make more conflicts).
However, it seems that you have a global var "env" in parser. I'm not sure how this will be merged. But I believe that our two groups will have overlapping when dealing with "[]".

parser.ts
```diff
@@ -232,38 +242,112 @@ export function traverseExprHelper(c: TreeCursor, s: string, env: ParserEnv): Ex
...
+          c.parent();
+          return { tag: "slice", obj: objExpr, index_e: endIndex };
+        }
+
+        // Start index or index
+        var startIndex = traverseExpr(c, s, env);
+        c.nextSibling(); // Focus on : or ]
+        if (s.substring(c.from, c.to) == "]") {
+          c.parent();
+          return { tag: "index", obj: objExpr, index: startIndex };
```
yiz136, line 285: We are set/dict group. We'll probably use your "index" expr in AST for our dictionary add (Because it's also a "[]" expression and we don't want to make more conflicts).
However, it seems that you have a global var "env" in parser. I'm not sure how this will be merged. But I believe that our two groups will have overlapping when dealing with "[]".

type-check.ts
```diff
@@ -261,6 +272,19 @@ export function tcStmt(env: GlobalTypeEnv, locals: LocalTypeEnv, stmt: Stmt<Anno
         throw new TypeCheckError(SRC, `field \`${stmt.field}\` expected type: ${JSON.stringify(fields.get(stmt.field).tag)}, got value of type ${JSON.stringify(tVal.a.type.tag)}`,
           tVal.a);
       return { ...stmt, a: { ...stmt.a, type: NONE }, obj: tObj, value: tVal };
+    case "index-assign":
```
yiz136, line 275: From set/dict group,
we'll have a huge conflict here (due to "[]" method)
A simple question, Why your annotation will have "a.type.tag" instead of "a.tag"?

Hackerry: Hi,
When we did our last merge, we incorporated code from [this PR](https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/16) which was already merged in the upstream repo at that time.

yiz136: No worry, we implement our "index" and "index-assign" first

PR #9, opened 2022-05-06
Design document for string.

williammm001:


designs/Strings-design.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Parser
+    - New case "String" on the traverseType
+    - New case "String" on the traverseLiteral
+    - New case "len" on the traverseExpr for CallExpression
+    - New condition for MemberExpression on traverseExpr when we are traversing the indexing
+
+- Type Checker
+    - New case "string" on typeCheckLiteral. Specifically, we need to check whether the escape sequences are valid.
+    - New case "indexing" on typeCheckExpr. Specifically, we need to check whether the expression that we perform indexing on is type "string".
```
jpolitz, line 229: And that the index is a number, and it should annotate the expression with `str`!

designs/Strings-design.md
```diff
@@ -0,0 +1,156 @@
...
+    - New case for typeCheckExpr - case "binexpr". For example, "abc" + 3 should throw a TYPE ERROR.
+
+- Compiler 
+    - New case "string" for codeGenLiteral, according to the length of the string we need to use the $heap pointer to allocate some space for the string
+    - New case "len" on the codeGenExpr for the builtin functions
+    - New case "Indexing" on the codeGenExpr for the indexing operations
+    - Modifications on the codeGenExpr when we applying concatenation of strings
```
jpolitz, line 242: I hope you don't have to modify `codeGen` at all and can handle this through lowering/flatten!

designs/Strings-design.md
```diff
@@ -0,0 +1,156 @@
...
+    
+## A description of the value representation and memory layout for any new runtime
+
+We would like to store the value of the string by (Extended)ASCII. Since ASCII is from 0 to 256, we would like to store a i32 to represent each character of a string into a continuous memory space and the start address of the string will be returned. Also, we will store the length of the string in the start address. In this way, when we reading a variable of string type, we can know where to stop (maybe useful for indexing).
```
jpolitz, line 254: This is a reasonable restriction for this project.

jpolitz (commented):
Good overall plan.

Please add tests where strings are arguments to functions, stored in class fields, etc! We need to make sure that strings interact OK with all existing features as well as the string-specific features.

What's your plan for adding `len`?

williammm001:
> Good overall plan.
> 
> Please add tests where strings are arguments to functions, stored in class fields, etc! We need to make sure that strings interact OK with all existing features as well as the string-specific features.
> 
> What's your plan for adding `len`?

IMO, we need to collaborate with the group implementing the builtin functions. Since the beginning address will be its length, we might need to add a new case that if the len() function is called and if there is a string inside. However, if other groups like the list group and the dictionary group also store the length in the memory, we do not need to differentiate the cases whether there is a string or a list inside the len().

designs/Strings-design.md
williammm001: Would you mind giving more details of that? I haven't get the point.

ZhiweiREN97: For example, in CodeGenValue, there is no "string" at all and in the CodeGenExpr, there is no "indexing" case. If we do not modify the codeGen functions in compiler.ts, how can we generate the WASM codes?

williammm001: After our discussion, we think that we can use the "store" statement in IR to store the string. However, one problem is that an assignment statement like "s1:str = "abc"" require we have the memory address of s1 which is not assigned in the original code. We think of one solution which is modify the env in TC. Would you mind telling us if it is a good idea?

PR #10, opened 2022-05-06
builtins design doc

jrdek:


LER0ever:
# DEPRECATED! DO NOT MERGE, SUPERCEDED BY #101 

~~WIP. DO NOT MERGE~~ DONE

Preview the design doc here: [`builtins-design.md`](https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/blob/66e0bd39fc744eb7491ae89b9c2daaf1ed273a85/designs/builtins-design.md)

## Summary
We plan to add / extend support for the following features next week:

- `print()`: arbitrary length arguments
- `int()` / `bool()`: type conversion
- `float`: data type initial support
- `import`: basic module and function import
- `math`: built-in math library as a simple PoC

## Breaking Changes
We will make breaking changes to the compilation pipeline next week, and probably needs instructors' approval before we do the refactoring, details in the proposal.

- `runner.ts`: additional AST transformation pass for import handling, potentially breaking for Error Reporting team due to line number mismatch
- `runner.ts`: additional HTTP requests to get the built-in library files
- `ast.ts`:  new type `ImportModule` and corresponding import Stmt

designs/builtins-design.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Import Handling
+We plan to place the import handling stage right after parsing before doing the type checking. We will use the information in AST, such as `imports` and function calls, to preprocess the input source code. For example, if we get the following source in its AST format, we will be able to recognize the import statement and the corresponding function calls depending on that import:
+
+```python
+from math import gcd, lcm
+print(gcd(12, 16), lcm(12, 16))
+```
+
+Then we rewrite the AST to include our library implementation, essentially replacing the original source file with the updated one below:
```
jpolitz, line 48: This is clever, but is a bad idea for a few reasons.

1. With transitive imports (a imports b imports c), what if `b` and `c` both import `d`? Won't that make two copies of `d`? What if `d` prints at the toplevel?
2. What if one of the imported functions relies on another function or global variable in the module as a helper, but isn't explicitly imported?
3. Consider duplicate classes with the above issue – will there be extra vtables created?
4. In the REPL, this will make each REPL entry be very long!
5. What if two _different_ modules define the same function and are imported into the same module?

(In C/C++, this is the reason for a whole bunch of nonsense with `#IFDEF`, etc.)

The strategy presented in the starter code, with `memory.wat` compiling to `memory.wasm` and then _linked_ through JavaScript might be a better one.

Then you could somehow communicate the updated _environment_ at the time of import (so from the compiler's point of view, the main thing that happens is a bigger starting environment and inserting some new WASM `import` statements).

designs/builtins-design.md
```diff
@@ -0,0 +1,201 @@
...
+        - unary: prints the argument
+        - n-ary: prints all arguments, separated by spaces
+
+## New functions, types, and/or files
+- `stdlib/math.py`
+    - factorial(n : int) -> int
+    - gcd(a : int, b : int) -> int
+    - lcm(a : int, b : int) -> int
+- `float`, as a type
+- various flavors of `import` statement
```
jpolitz, line 104: Is there a “most basic” one you can do for week 1, maybe `from x import y, z` since it doesn't involve a new meaning for the `.` operator?

jpolitz (commented):
Thank you for this detailed description, it really helped me understand and is well-written!

I want you to rethink the copying code strategy for the reasons described inline – this proposal is more or less C/C++ `#include`, which has some shortcomings in our case (some can be solved, the REPL is really hard to do this way). If you really want to pursue it, I'm game to try, but I warn against it.

Also, what's your proposed code layout? Like where will library files go, will you add to the makefile, etc?

LER0ever:
Implementation for Milestone 1 moved to #46

PR #11, opened 2022-05-06
For loops & Iterators 

Ananya273:


design/design.md.md
```diff
@@ -0,0 +1,114 @@
...
+for i in range():
+  print(i)
+
+Expected Output : TYPE ERROR: expected 2 parameters instead got 0
+
+TestCase 2: Parameters of range should be integers and start <=end. Else, throw typecheck error (design decision)
+
+Input
+i:int = 0
+for i in range(10,1):
```
jpolitz, line 23: Some of these range examples have 2 arguments and some only 1. How will that work?

design/design.md.md
```diff
@@ -0,0 +1,114 @@
...
+  print(i)
+  
+
+Expected output: 0 1 3
+
+Testcase 10: Type checking iterator and values list
+
+Input 
+i:bool = True
+for i in range(4):
```
jpolitz, line 104: One thing I noticed is that there aren't any `range` examples with a variable or non-number as an argument to range. I think a test of a function/method that uses `range` on the argument would be helpful (e.g. write `sum(n : int) -> int` that adds up `1` to `n` in a loop and returns it).

design/design.md.md
```diff
@@ -0,0 +1,114 @@
...
+   print(i)
+   i = True 
+   print(i)
+
+Expected output: Type Error expected bool, got int
+
+Changes to AST.ts include 
+
+export type Stmt<A> =
+  | {  a?: A, tag: "for", iterator: Expr<A>, values: Expr<A>, body: Array<Stmt<A>> }
```
jpolitz, line 114: I'm a little confused about why `iterator` is an `Expr<A>`. That would mean I can write things like

```
for 1 + 2 in range(0, 10):
  pass
```

Which doesn't make any sense. Should iterator be an id?

design/design.md.md
```diff
@@ -0,0 +1,114 @@
...
+
+Expected output: 0 1 3
+
+Testcase 10: Type checking iterator and values list
+
+Input 
+i:bool = True
+for i in range(4):
+   print(i)
+   i = True
```
jpolitz, line 106: Good to test this typing!

jpolitz (commented):
I'd like to see a little more about the overall plan for the compiler, along with some tests that combine loops (in the restricted form you suggest) with some other language features.

design/design.md.md
akchekur: Sure, thanks for the suggestion. Will include this in our test cases.

design/design.md.md
akchekur: We were planning to implement basic function overloading (using different names for the same function based on the parameters as of now) until the code from other teams is available.

design/design.md.md
akchekur: yes, the iterator should be an "id" and not expression. Thanks for the comment.

PR #12, opened 2022-05-06
Inheritance Design Documentation

anuj-ahuja:


designs/inheritance-design.md
```diff
@@ -0,0 +1,236 @@
...
+	l = Link().new(5, Link().new(13, Empty()))
+	print(l.sum()) # prints 18
+	
+# File Changes
+
+### type-check.ts
+
+  1. Add a new function to check if a class is a sub class of another class.
+
+### compiler.ts
```
jpolitz, line 234: I think some more detail might be needed here with the presence of the IR.

For example, which of these operations will be represented just by the AST?

How will the compilation of method calls be split between lowering and the IR?

Agree that the method table needs to be part of the last step.

ast.ts
```diff
@@ -12,7 +12,7 @@ export type Parameter<A> = { name: string, type: Type }
 
 export type Program<A> = { a?: A, funs: Array<FunDef<A>>, inits: Array<VarInit<A>>, classes: Array<Class<A>>, stmts: Array<Stmt<A>> }
 
-export type Class<A> = { a?: A, name: string, fields: Array<VarInit<A>>, methods: Array<FunDef<A>>}
+export type Class<A> = { a?: A, name: string, fields: Array<VarInit<A>>, methods: Array<FunDef<A>>, super: Array<String>}
```
jpolitz: I notice there's a whole array of super positions here – is the plan to support multiple inheritance eventually?

designs/inheritance-design.md
```diff
@@ -0,0 +1,236 @@
...
+
+	l : List = None 
+	l = Link().new(5, Link().new(13, Empty()))
+	print(l.sum()) # prints 18
+	
+# File Changes
+
+### type-check.ts
+
+  1. Add a new function to check if a class is a sub class of another class.
```
jpolitz, line 232: There might be some other details here, too, right? Like getting the method type for a class that doesn't implement it.

designs/inheritance-design.md
```diff
@@ -0,0 +1,236 @@
...
+		def __init__(self: B): pass
+	
+	class C(B):
+		c: int = 5
+
+		def __init__(self: C): pass
+			
+	l : C = None
+	l = C()
+	print(l.a) # Print 1
```
jpolitz, line 197: Glad you have fields tests!

designs/inheritance-design.md
```diff
@@ -0,0 +1,236 @@
...
+
+  1. Add a new function to check if a class is a sub class of another class.
+
+### compiler.ts
+
+ 1. The compiler will create a WASM table having the all the method calls for each class.
+ 
+ 2. Add a data structure that maps each class to its start index in the vtable where the methods for the class are stored. We also need to store this index as an additional i32 field on each object we create so that we can lookup the correct start index for each object.
+
+ 3. Add a data structure that maps each class-method pair to an offset.
```
jpolitz, line 240: And presumably class-field? Can that happen at the lowering level?

jpolitz (commented):
Nice work. Some detail comments about implementation strategy to nudge you to think about some complexities. Generally look solid.

Abhishek2304:
Add passing objects to functions tests?

Abhishek2304:
Maybe need to handle subtyping: Make sure to use the provided functions for type evaluation/subtyping.

Abhishek2304:
Implement a to_string() function for the classes (for passing to print used by front-end)

Abhishek2304:
Write tests in new test file (like inheritance_test.ts)

ast.ts
anuj-ahuja: Yes

designs/inheritance-design.md
anuj-ahuja: Yes, it would be better to have these at the lowering level to keep the compilation process smooth. We have reflected this in the updated design document.

designs/inheritance-design.md
anuj-ahuja: Updated in the design document.

designs/inheritance-design.md
anuj-ahuja: Updated in the design document.

PR #13, opened 2022-05-06
Testing design

jasoonn:


designs/testing-design.md
```diff
@@ -0,0 +1,241 @@
+Testing
+===
+##  Integrated test cases
+Note : We will come up with more integrated test cases next week and make sure that the syntax of our test cases align with the design documents of other groups.
```
jpolitz, line 4: Yay!

designs/testing-design.md
```diff
@@ -0,0 +1,241 @@
...
+```
+Expected output:
+```
+False
+True
+```
+
+## Changes we want to make
+#### changes in tests/
+- Add more integrated test cases
```
jpolitz, line 239: These are great tests!

I would suggest some tests that are a _little_ simpler than these, that focus on combining just, say, two features.

For example, just using all of the different kinds of new values as function arguments and class fields is a good way to make sure their memory allocation strategy/value representation is working well.

Or just doing destructuring assignment on the result of a comprehension (or a for loop on the result of a comprehension) will make sure those are tested together.

So think of 2-feature combinations.

All the ones you have are excellent and important, too!

jasoonn: Got it. We will first focus on combining two features tests! Thank you!

PR #14, opened 2022-05-07
Comprehensions design

pratyush82:
initial designs and test cases


designs/comprehensions-design.md
```diff
@@ -0,0 +1,105 @@
...
+a : int = 5
+b : List = None
+b = [i for i in a]
+print(b)
+```
+This should output ```Error: variable a is not an iterable```
+
+## Code changes
+
+We plan to have an external class Iterable for now with corresponding ```next()``` and ```hasNext()``` methods. Every time a list comprehension will be done, a new instance of this class will be created and used to assign values to the new list variable. We want to create a seperate file to construct the range class under the directory models. We can talk about the range class in more details after we have a discussion with the lists group and once we start our implementation.
```
jpolitz, line 286: My suggestion is to not make `range` be a built-in. Rather, for the tests you write in this iteration, just put a class with a `next` and `hasnext` method in every test you write (you can use a helper to concatenate it and avoid repeating yourself). That way you don't have one hardcoded class that works.

ast.ts
```diff
@@ -38,6 +39,17 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "list-construct", items: Array<Expr<A>> } // implemented by the list team.
```
jpolitz: Let's let the list team add it, then, because otherwise we'll have conflicts.

ast.ts
```diff
@@ -7,6 +7,7 @@ export type Type =
   | {tag: "none"}
   | {tag: "class", name: string}
   | {tag: "either", left: Type, right: Type }
+  | { tag: "list", listitems : Array<Type>}
```
jpolitz: I'm confused by this type. Why will lists have a variable number of _types_ of elements, and why is this part of comprehensions, not lists?

designs/comprehensions-design.md
```diff
@@ -0,0 +1,105 @@
...
+a : List = None
+a = [i for i in range(10)]
+print(a)
+```
+This should output ```[0,1,2,3,4,5,6,7,8,9]```
+
+```
+a : List = None
+j : int = 9
+a = [j for i in range(5)]
```
jpolitz, line 210: Note that some of what's required here is similar to closures (in order to get references to `j` to work out)

designs/comprehensions-design.md
```diff
@@ -0,0 +1,105 @@
+# Comprehensions Team Design Doc
+
+## Test cases
+
+```
+a : List = None
```
jpolitz, line 201: I would recommend _not_ using `List` types here. First of all, `List` isn't enough to know the type of the _contents_. Second, you're dependent on the lists group to run anything at all.

I would recommend writing your tests in terms of `()` comprehensions so that you're in total control of the generator/iterator representation (as you discuss below, by creating new classes).

jpolitz (commented):
Good tests, but see my comments about getting started _without_ relying on lists.

ast.ts
shreyasaha1997: We do agree with this. The only reason we have mentioned it here is so that we could construct the ast structures while parsing. Are you suggesting that we just use the next() and the hasNext() methods for now?

ast.ts
```diff
@@ -39,7 +39,18 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "index", obj: Expr<A>, index: Expr<A> }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
-  | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "construct", name: string, arguments: Array<Expr<A>> }
```
yalhessi: Why are you adding arguments to `construct` here? This is a much bigger change than you may realize. We've discussed this in class a few weeks ago but one of the reasons ChocoPy goes with no arguments to construct is so that you're not allocating objects that are arguments in the middle of allocating other objects, and therefore making bookkeeping really complicated.

debugstart.ts
```diff
@@ -1,23 +1,86 @@
 import { parse } from "./parser";
```
yalhessi: Avoid including debugging-related files/changes.

ast.ts
```diff
@@ -54,3 +65,7 @@ export enum UniOp { Neg, Not };
 export type Value =
     Literal
   | { tag: "object", name: string, address: number}
+
```
yalhessi: Whitespace!

ast.ts
```diff
@@ -39,7 +39,18 @@ export type Expr<A> =
...
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
-  | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "construct", name: string, arguments: Array<Expr<A>> }
+  // ast for list-comp
+  | {  a?: A, tag: "list-comp", left: Expr<A>, elem: Expr<A>, iterable: Expr<A>, cond?: Expr<A>, iterable_cond?: Expr<A>, body?: Array<Stmt<A>> }
+
+  // Possible Typechecks here ::
+  // 1. iterable.a.tag = list
+  // 2. cond is optional and cond.a.tag = bool and cond.tag = "if"
+  // 3. elem.tag = id
```
yalhessi: Why have elem be an Expr? Why not a string? Or a LHS?

ast.ts
```diff
@@ -39,7 +39,18 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "index", obj: Expr<A>, index: Expr<A> }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
-  | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "construct", name: string, arguments: Array<Expr<A>> }
+  // ast for list-comp
+  | {  a?: A, tag: "list-comp", left: Expr<A>, elem: Expr<A>, iterable: Expr<A>, cond?: Expr<A>, iterable_cond?: Expr<A>, body?: Array<Stmt<A>> }
```
yalhessi: What's body here?

designs/comprehensions-design.md
```diff
@@ -0,0 +1,216 @@
...
+5. simple comprehension output using iterable class methods
+```print([e for e in range(1,5)])```
+Output: ```['1', '2', '3', '4']```
+
+6. simple comprehension output using iterable class methods and expr values
+```print([f*3 for f in range(1,5)])```
+Output: ```['3', '6', '9', '12']```
+
+7. simple comprehension output with if condition
+```print([g for g in range(1,5)])```
```
yalhessi: Where is the if?

designs/comprehensions-design.md
```diff
@@ -0,0 +1,216 @@
...
+```
+ a : List = None
+a = [i for i in range(5,10)]
+print(a)
+```
+In case of the second test case, we skipped it because as suggested by proffesor, we plan on using the implementation by the lists team. For this week, we only worked on iterables using the range class.
+
+## Limitation so far (plan on fixing in week 8)
+
+1. We made an assumption that the list comprehension expressions will only be used globally, and not locally inside any function.
```
yalhessi: What changes do you need to make to generalize the context of comprehensions?

designs/comprehensions-design.md
```diff
@@ -0,0 +1,216 @@
...
+2. The explicit range class inside each of our test cases has a class member as curr. If we have an expression as [a for a in range(10)], then we assign a to range.curr (curr refers to the current element in the iteration). However, we are assigning a as a global variable, instead of giving it a local scope(ie, scope of variable a should only be within the comprehension). Because of this, every time we want to use a different comprehension expression, we are using a new counter variable.
```
yalhessi: Why did you go with adding the element as a global variable and not local? What would you need to do to fix it because that's definitely unexpected behavior?

designs/comprehensions-design.md
```diff
@@ -0,0 +1,216 @@
...
+```
+This should output ```Error: variable a is not an iterable```
+
+## Code changes
+
+We plan to have an external class Iterable for now with corresponding ```next()``` and ```hasNext()``` methods. Every time a list comprehension will be done, a new instance of this class will be created and used to assign values to the new list variable. We want to create a seperate file to construct the range class under the directory models. We can talk about the range class in more details after we have a discussion with the lists group and once we start our implementation.
+
+In the ast.ts file,  we plan to add a new Type for lists:
+```
+{ tag: "list", listitems : Array<Type>}
```
yalhessi, line 290: This really needs to be added as soon as possible.

lower.ts
```diff
@@ -113,6 +113,27 @@ function flattenStmt(s : AST.Stmt<Type>, blocks: Array<IR.BasicBlock<Type>>, env
     // ]];
   
     case "expr":
+      // list-comp changes
+      if(s.tag == "expr" && s.expr.tag == "list-comp")
```
yalhessi: Doesn't the case already check that the tag is "expr"?

type-check.ts
```diff
@@ -184,6 +253,16 @@ export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<n
...
       const tExpr = tcExpr(env, locals, stmt.expr);
+        
+      // code to generate IR for list-comps
+      if(tExpr.tag === "list-comp")
+      {
+        return tcListComp(env,locals,stmt,tExpr);
+      }
+      if(tExpr.tag === "builtin1" && tExpr.arg.tag === "list-comp")
+      {
+        return tcListComp(env,locals,stmt,tExpr.arg);
```
yalhessi: What's this case?

type-check.ts
```diff
@@ -313,16 +392,67 @@ export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<n
       } else {
         throw new TypeError("Undefined function: " + expr.name);
       }
+    case "list-comp":
+      // check if iterable is instance of Range class
+      const iterable = tcExpr(env, locals, expr.iterable);
+      if (iterable.a.tag === "class" && iterable.a.name === "range"){
```
yalhessi: I would definitely generalize this to things other than `range`. We suggested considering any class that implements a `next` and `hasnext` as a possible interface.

yalhessi (changes requested):
Nice work!

Seems like you've made some good progress here. It's a bit hard to judge how fragile your implementation is here without spending some more time writing tests. Here are some things I would like to see:
1) Write more tests that cover different uses of the comprehensions. You seem to mostly assign the comps to prints or leave them unbound. Try assigning them to variables. Maybe in if statements. Test nested comprehensions. Range with non-literal expressions. All of these could be cases where your implementation breaks. 

2) You'll need to rework your implementation a bit to make sure `construct` doesn't take arguments (unless you really thing this is critical in which case you should talk to other groups, mainly MM, to see if nested construction could be handled safely.

ast.ts
shreyasaha1997: This has been removed.

debugstart.ts
shreyasaha1997: This has been removed.

ast.ts
shreyasaha1997: We agree with you. We reverted back these changes.

ast.ts
shreyasaha1997: These comments are actually part of our design decisions from Week 1 of the project. Although the comments are now removed, our though process was as follows: [x for a in range(5)], where x is represented by elem. Now this x can be anything ranging from a variable or a string or some mathematical value (involving unary or binop) or a function call. And every one of these cases comes under expression.

ast.ts
shreyasaha1997: This extra parameters here have been removed.

designs/comprehensions-design.md
shreyasaha1997: This has now been fixed.

designs/comprehensions-design.md
shreyasaha1997: This limitation has now been fixed and I have described the necessary changes in the comment below.

designs/comprehensions-design.md
shreyasaha1997: We have fixed this. In the previous version of our code, we were creating range objects in the type check file that were global in nature. We have a more generalized code in type check now, that we are flattening in lower.ts. Th limitation is no longer valid and has been updated.

designs/comprehensions-design.md
shreyasaha1997: Regarding this, in week 1 we got a review which advised us to wait for the lists team to make this change. We plan on utilizing the change made by the lists team as part of our next milestone.

lower.ts
shreyasaha1997: This has been fixed.

type-check.ts
shreyasaha1997: This case is no longer required and has been updated.

type-check.ts
shreyasaha1997: We plan on fixing this hardcoding as part of our next milestone. Please let us know if that is not okay, we will try to resolve this.

shreyasaha1997:
Overall comments:

1. The only reason we did not assign the comprehension expressions to any variable and have left them unbound, is because we wanted to use the list variable type implemented by the lists team. The lists team is also responsible for storing the list elements in memory. This was advised to us in the first week review.
2. We do agree that constructors should have only one element which is the self object. We have reverted our changes in this particular aspect.

type-check.ts
pratyush82: This has been fixed. Now the type-check does not check for a class named "range". Instead, it just checks if the class (can be any name) has the "next" and "hasNext" methods. The iterable can now be an instance of any class with those two methods.

PR #16, opened 2022-05-07
Error Reporting Design

madun1999:


designs/error-reporting-design.md
```diff
@@ -0,0 +1,341 @@
...
+
+Here's an overview of the changes in code:
+
+```ts
+Annotation = {
+    type?: Type,
+    fromLoc: Loc,
+    toLoc: Loc,
+    fullSrc: string
+}
```
jpolitz, line 56: This will involve changing lots of code that sets the `a` field, right? (I don't know that I have a great way for you to get around that, just making sure).

designs/error-reporting-design.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+## Functions to Add
+### Parsing
+During parsing, we will need to record the starting and ending location of each AST node. We need to figure out how to get line and column number out of lezer or count them ourselves. We may add `annotateLocation(ast, lezer_cursor)` function that calculates location information from a lezer cursor and add to the annotation field of the AST node.
+
+### Type Checking
+We will add a collection of functions, such as `binaryOperatorTypeError(ast)` that helps format the error message using the annotation in the AST node.
+
+### Compiler
```
jpolitz: Oh, cool! If you do this, talk to the memory mangement group. They might want this.

designs/error-reporting-design.md
```diff
@@ -0,0 +1,341 @@
...
+a = True
+```
+
+Should report the following static error:
+
+```
+TypeError: assignment value should have assignable type to type `int`, got `bool` on line 1 at col 4
+
+a = True
+    ^^^^ attempt to assign type `bool` to type `int`
```
jpolitz, line 180: So cool!

designs/error-reporting-design.md
```diff
@@ -0,0 +1,341 @@
...
+```
+#### Out of memory
+If the program is out of memory, the following runtime error should be reported:
+```
+Traceback:
+  Toplevel, line 10
+
+OutOfMemoryError: Out of memory
+```
+We might improve this reporting behaviour based on the kind of errors that the memory manager reports.
```
jpolitz, line 426: Nice. Yeah `alloc` should be able to do this nicely.

jpolitz (commented):
Seems like a great start.

I think coordinating how to get a consistent API/structure for annotating source locations without messing up others will be important. Think about what the code edits will look like for that, and how folks who are adding _new_ code will work within that structure.

designs/error-reporting-design.md
madun1999: We can get around that by keeping the current `a` field as is and adding a new field to every AST node that stores the location info we want. 
But we thought that maybe other groups also want to annotate the AST with relevant information, and we might as well put everything in one place. It makes changing the annotation down the line easier. 
The decision depends on if we want to rewrite the previous type checking codes that uses the `a` field.

PR #17, opened 2022-05-07
memory management milestone

tgujar:


designs/memory-management-design.md
```diff
@@ -0,0 +1,275 @@
...
+
+x: Rat = None
+x = Rat()
+```
+**Expected:**
+Let `o` be the object referred to by variable `x`
+```
+assert number of references of o is 1
+assert size of o(in bytes) is 4
+assert type of the fields in o is [value]
```
jpolitz, line 21: Nice structure for these tests.

designs/memory-management-design.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Changes to IR
+
+- ### Addition of `ref` value
+```
+export type Value<A> =
+...
+| { a?: A, tag: "ref"}
+```
+Currently there is no way of differentiating between pointers and values stored in an object. As seen in test case *8: Anonymous object deletion*, we need to follow references stored in an object recursively so that we can update their respective reference counts. This is not possible without knowing which field represents a reference/pointer.
```
jpolitz, line 250: I'm confused about this. There's no runtime information on `ref`, so it's not clear what this will compile to/be stored as. Can't you find out which fields are references/pointers by looking up the class and using type information? If it's bool/int then it's not a pointer, if it's class then it is? Statically I think this isn't a problem.

designs/memory-management-design.md
```diff
@@ -0,0 +1,275 @@
...
+
+- ### Changes to `alloc` expression
+```
+export type Expr<A> =
+  ...
+  | {  a?: A, tag: "alloc", amount: Value<A>[] }
```
jpolitz, line 256: So it looks like the change here is that this will allocate a fixed list of values (always). Is `amount` the right name for it, then?

Is the idea that in code generation you'll give some kind of information to the `alloc` function about whether each of these is a `ref` or not? How will the code generation work here, because the `alloc` function can't be variable-arity.

I definitely agree that you need some type information to represent this to alloc. Could you make use of the `a` field to give you the type of the contents, and set the metadata that way?

Here's the case I'm concerned about – the `lists` group presumably will want to implement some kind of array list. This involves creating fixed-length arrays with computed sizes and moving elements around. That means the _number_ of elements isn't statically known. How should that group accomplish their task with only fixed-length alloc?

designs/memory-management-design.md
```diff
@@ -0,0 +1,275 @@
...
+```
+ To follow references and to deallocate memory on the heap, we need the size and the type information of an object which is stored on it. This expression, allows the memory manager to understand how many bytes are to be allocated/ deleted when the object is garbage collected and what will be the type of data in a field.
+
+- ### Remove BigInt type
+We feel that the implementation of `bigint` type would probably be as a contiguous array of `int` and could just be made into an `object`. This relates to our implementation since we would not have to account for a special case.
```
jpolitz, line 261: Ah, interesting. So you're proposing to not have `bigint` at the IR level and just make it the group's problem to figure out how to alloc it. I buy that.

designs/memory-management-design.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Added functions/ data types
+New data types are added only to IR and are same as described above.
+
+## Value representation and memory layout
+
+![Memory layout with metadata](./images/memory-metadata.drawio.png)
+
+The type of a field is denoted by a single bit. For the memory manager we feel that it does not really matter what the type of an field is, as long as we are able to differentiate between data and references to data. The current representation that we plan uses 32 bits to represent types in an object, which puts an upper bound on the number of fields in an object to 32. If needed this number can be easily increased, assigning more bits to storing types in the metadata.
```
jpolitz, line 270: I think this is a reasonable restriction for now.

jpolitz (commented):
See my challenge to you about how to implement array lists with only fixed-length allocations. (Maybe there's two allocs – computed-length-but-homogeneous and fixed-length-heterogeneous?)

designs/memory-management-design.md
tgujar: From our conversation today, I understand that we can look up the type of an object by looking at the `a` field in `alloc` and then determining if a field is a `reference` or a `value` by looking up the class information on the fields.


`amount` field can then be used to denote the number of elements of type `a` which are to be allocated in a contiguous fashion on the memory. This would also solve the issue with dynamic-size allocation. I think this is what the original IR was going for and we might have misinterpreted some stuff.

I guess we could keep the IR same as original in this case, and get all info we need as below
- Size info - determined by the `amount` field
- Field info(reference or value) - determined from class `fields`, since `VarInit` has type info

PR #18, opened 2022-05-07
Destructuring Assignment Design

zhuo-zhi:
Initial Design Doc and test cases.


jpolitz (commented):
Thank you for your clear description!

The design proposed here is too restrictive, I think, to be usefully extended in the future.

In particular, I want you to come up with a design that can handle (roughly) this case:

```
def f(n : int) -> int:
  a = range(0, n)
  v1, v2, v3, v4, v5 = a
  return v1 + v2 + v3 + v4 + v5
f(5)
f(3) # throws a dynamic error "too many values to unpack"
```

In the getting-started Piazza post, I suggested introducing something like `range` so you can have an iterator to work with. In general, the right-hand side can be any iterable value (which includes lists, tuples, etc), which means that, grammar-wise, it could be _any expression_ in the AST.

In the above example, _we cannot statically check_ that the destructuring assignment has the right number of elements. Instead, we need to generate code that will check lengths and do the appropriate iteration and assignment at runtime.

zhuo-zhi:
Above new commits is the code for project milestone 1 on May 11.

pancerZH:
> Above new commits is the code for project milestone 1 on May 11.
Sorry the milestone is above commit id[08d46c2] (included)

PR #19, opened 2022-05-07
Io file chaote

Chao-Te:
Finish and upload the design doc for compiler A: I/O - file

### DIFF
--- a/designs/io-file.design.md
+++ b/designs/io-file.design.md
@@ -17,22 +17,22 @@ export type openmode = {
 }
 ```
 ### TODOs
-1. Basic methods we might need to support for file objects: 
-    -  Open(int mode m) -> file: open a file in mode m.
-    -  read(int i): read i bytes from file. Currently will return anything. After the implementation of string and bytes, we will add the return type.
-    - write(): write "test" to the file.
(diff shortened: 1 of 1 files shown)


jpolitz (commented):
I think there are some key things I'd change about this design.

It's not clear from this document where the files are actually stored. This could be as simple as a dictionary in javascript (to start) or something more sophisticated like the [FileSystem API in browsers](https://developer.mozilla.org/en-US/docs/Web/API/File_and_Directory_Entries_API/Introduction). But loading and reading the file from _somewhere_ is important, and needs to be specified.

I agree that some simplifications need to be made to get started. I think your restriction of `open` to only static strings for now might work well (which would require a new AST type for `open` expressions). However, I don't really understand how read and write are working. Here's a concrete suggestion:

```
read() -> int # returns the byte at address 0 in the file
write(int : c) -> int # writes the byte at address 0 in the file
```

You could go one level more sophisticated with e.g. `seek` to a location in the file. But this keeps all the operations working with ints, so you aren't waiting for the strings group, and makes you think through all the pieces of loading/saving to files somewhere from an API that you provide.

How does that sound?

PR #20, opened 2022-05-07
bignums design

taipeiyuka:


designs/Bignums-design.md
```diff
@@ -0,0 +1,104 @@
+# Compiler A: Bignums - design
+
+## AST, IR, and Built-in Libraries
+- We will change `print` in the `built-in library` to print bignums from the heap.
+- We plan to modify the `built-in library` to have Javascript functions support all the unary and binary operations of bignums. This way we can first support the behavior of Python, and then improve the efficiency by writing the functions in WASM.
+
+## New Functions, Datatypes, and/or Files
+- We will change `codeGenValue` to allocate bignums in the memory. This will store the bignum in memory and leave the address on the stack.
+- We will modify `codeGenUniOp` and `codeGenBinOp` to support operations.
```
jpolitz: How will you make sure that you don't break any other code that relies on the existing op behavior?

designs/Bignums-design.md
```diff
@@ -0,0 +1,104 @@
...
+- We will add `codeGenMemoryIndex` to let strings or lists access memory with bignums as indices. This function will take the base address and the bignum index, and return the memory address you want to access.
+
+## Value Representation and Memory Layout
+
+- A bignum stores the underlying number in digits of 2^30-based.
```
jpolitz: Why 2^30 and not 2^32 or 2^31?

designs/Bignums-design.md
```diff
@@ -0,0 +1,104 @@
+# Compiler A: Bignums - design
+
+## AST, IR, and Built-in Libraries
+- We will change `print` in the `built-in library` to print bignums from the heap.
+- We plan to modify the `built-in library` to have Javascript functions support all the unary and binary operations of bignums. This way we can first support the behavior of Python, and then improve the efficiency by writing the functions in WASM.
```
jpolitz: I like this idea, no need to do this in WASM up front.

designs/Bignums-design.md
```diff
@@ -0,0 +1,104 @@
...
+
+## New Functions, Datatypes, and/or Files
+- We will change `codeGenValue` to allocate bignums in the memory. This will store the bignum in memory and leave the address on the stack.
```
jpolitz: How will this interact with e.g. alloc/load/store so the memory management group knows what's going on with bignums? Remember that we don't have access to `$heap` anymore from generated code, only `memory.wat` has access to that.

designs/Bignums-design.md
```diff
@@ -0,0 +1,104 @@
...
+- We will modify `codeGenUniOp` and `codeGenBinOp` to support operations.
+- We will add `codeGenMemoryIndex` to let strings or lists access memory with bignums as indices. This function will take the base address and the bignum index, and return the memory address you want to access.
```
jpolitz: This is interesting – is the expectation that those groups would call this function somehow, or is there a new IR form for it? When does this get called?

jpolitz (commented):
Thanks for your detailed proposal!

There are a few important questions inline in the doc, but one big question that isn't addressed: how does transitioning between little numbers and bignums work? For example, what happens if you evaluate `4294967291 * 4294967291` (4294967291 is 2^32 - 5).

What happens if you evaluate `42949672910 / 4294967291`?

These are both cases that mix numbers that could be represented as 32-bit integers and numbers that cannot. How does the system represent these? Are all numbers bignums, or is there some way to convert?

designs/Bignums-design.md
taipeiyuka: Does "interact with e.g. alloc/load/store" mean we have to know the size of the bignum we want to alloc before run-time? Shouldn't there be some way to reallocate a larger space in run-time like vectors in C++?

taipeiyuka: We could make function calls to alloc/load/store in `memory.wat` to make runtime allocation.

taipeiyuka: We could make function calls to alloc/load/store in `memory.wat` to make runtime allocation.

PR #21, opened 2022-05-07
Generics - design

uvashishtha:
## This PR outlines the design for supporting generics.

### DIFF
--- a/designs/generics-design.md
+++ b/designs/generics-design.md
@@ -1,14 +1,23 @@
 ## Test Cases
 ---
 
+Generics are a programming language feature that allow classes and functions to
+abstract over the specific types of data that they work with. This is done by
+parameterizing classes and functions over the types that they can work with.
+
 1. **Basic Type Variable Declaration**
 
 ```python
 T = TypeVar('T')
 ```
 
-This should add a type variable T, with the descriptive name 'T', to the global typing environment. 
-This name can be potentially be used for descriptive error messages from the compiler.
+A type variable can be introduced using a TypeVar definition. New type variables
+can be introduced only in global scope and can used across any number of generic
+class/function definitions.
+
+The first parameter to TypeVar can be used to provide a more descriptive name
+for the type variable and can potentially be used for more descriptive error
+messages from the compiler for example.
 
 2. **Class Definition with a Type Variable**
 
@@ -18,18 +27,31 @@ T = TypeVar('T')
 class Box(Generic[T]):
     pass
 ```
-This should typecheck the class **Box** and add this to the global state of classes with a generic type parameter, T.
 
-3. **Object creation using Type Variable instantiated with a Primtive Type**
+A generic class can be defined by inheriting from the special base class Generic
+along with a defined type variable. Any other class that this class
+inherits from in the inheritance hierarchy can be specified after this.
+
+A generic class is type-checked for all possible allowed instantiations of its
+type variables independent of the instantiations actually made in the program.
+
+3. **Object creation using Type Variable instantiated with a Primitive Type**
 ```python
 T = TypeVar('T')
 
 class Box(Generic[T]):
     pass
 
-b: Box[int] = Box()
+b: Box[int] = None
+b = Box()
 ```
-This should 
+
+An object of a generic class can be created by instantiating the type parameter
+with a primitive type in the type annotation for the object's variable declaration.
+
+After type-checking and before lowering a copy of each generic class is made for
+each type that its type variables are instantiated with over the course of the program.
+
 
 4. **Object creation using Type Variable instantiated with a User-Defined Type**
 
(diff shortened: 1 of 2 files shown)


designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+```
+
+A type parameter of a generic class can also be instantiated with a user-defined
+class.
+
+5. **Type Variables and Generic Classes with Constraints**
+
+```python
+T = TypeVar('T')
+R = TypeVar('R', int, bool)
```
jpolitz, line 78: I think we could leave constraints for a future milestone – there's a lot to do just for the unconstrained version.

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+b : Box[int] = None
+b = Box()
+print(b.f) #prints 0 (zero value of int)
+b.f = 10
+print(b.f) #prints 10
+```
+
+Generic classes can have fields with the type parameter as a type annotation. Since
+fields need to always be initialized but we don't know the actual type of the field, generic
+fields need to be initialized with the special __ZERO__ literal value that automatically uses
```
jpolitz, line 110: Ah fascinating! Nice to leave this until later. Special symbols is an interesting idea. I wonder if we need a better convention for them, but this is fine for a start.

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+```python
+T = TypeVar('T')
+
+class T(): # TypeError: Duplicate identifier in the same scope.
+    pass
+```
+
+Type Variable names cannot overlap with class, function or variable names. This
+is because in some contexts we need to be able to unambiguosly determine if an
+identifier is a type variable.
```
jpolitz, line 148: Good choice

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+An independent function can also be generic. The arguments and return type can use
+any of the type variables that are declared in the program.
+
+11. **Joe's blessed test case from class**
+```python
+T= TypeVar('T')
+class C(Generic[T]):
+    t: T = __ZERO__
+    def f(self: C[T], other: C[T]):
+        other.t = self.t
```
jpolitz, line 185: Right, and here the type being annotated correctly on `self`/`other` in the end will be important for typed loads and stores with memory management/offsets.

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+```
+
+---
+
+## Proposed changes to the AST
+
+```typescript
+export type Type =
+  ...
+  | { tag: "class", name: string, params: Array<Type> }
```
jpolitz, line 204: Noting that this will conflict the inheritance group (both in parsing and in representation). Need to get that right.

jpolitz: Maybe make params `?`-optional because it's annoying to create empty arrays for most classes?

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+
+A class can have multiple type parameters. These type parameters can be independently
+instantiated to different types based on their own constraints.
+
+10. **Function that uses Type Parameters**
+```python
+T = TypeVar('T')
+
+def id(t: T) -> T:
+    return t
```
jpolitz, line 173: How do we call `id`? Do we need any special syntax for that?

designs/generics-design.md
```diff
@@ -0,0 +1,403 @@
...
+  typevars: Map<string, TypeVar>,
+}
+```
+
+### Type Hierarchy and Comparison
+
+The below functions need to be modified to take into account
+type parameters and generic types. With type parameter
+constraints, these could get pretty complex and we have not
+fleshed out the exact details completely at the moment.
```
jpolitz, line 364: I would ignore constraints and do exactly equality on type parameter names. The inheritance group will already add nontrivial subtyping, let's not give ourselves too much to take on.

jpolitz (commented):
This is all great. I think it's worth sequencing your work a little bit – maybe leave out functions and just focus on classes to start (or vice versa) to get the monomorphization pass down so things can actually run. Maybe just do fields + classes first to get the interesting memory offset issues done, then bring in methods and functions later (so error if tyvars are used in those positions).

designs/generics-design.md
abhishekc-sharma: Calling it does not require anything special in our current design. The type-checker will notice that `T` refers to a type-variable and mark the function as generic. The monomorphization will then find calls to such functions and duplicate the function with the proper types.

designs/generics-design.md
abhishekc-sharma: From a consistency perspective we felt it was better to always have params and treat normal class types as those without parameters. The annoyance of having to deal with this will hopefully be hidden behind the `CLASS` utility function taking an optional parameter for params and that the entire code-base uses that utility function when needing to create class types.

designs/generics-design.md
jpolitz: What will happen if I call `id(None)`?

abhishekc-sharma: Our current thinking is that this would cause the function to be monomorphized for the "none" type and return None.

abhishekc-sharma:
Is there a plan to actually merge this PR ? Asking because we created our other branches off of this one and were holding off merging so that this PR doesn't get updated.

PR #23, opened 2022-05-07
frontend design

Gandalfthegreat:


design/frontend-design.md
```diff
@@ -0,0 +1,147 @@
...
+            e: none
+        }
+```
+5.  Add one HTML component <select></select>:
+It would be much more convenient for users to begin with some sample code. We could provide a HTML component listing the example codes. For example, the options of the select could be "example code with class", "example code with closure". Once we click one option, the code would be loaded to the left box on the screen.
+
+6.  Switch the select component:
+When we change the option of the select, the example code would be overwritten. When we clicked the "example code with closure" option, the code in the left box would be overwritten.
+
+7. "Clear" button:
```
jpolitz: Can we put the buttons in better places, too? Like think about nice alignment for them?

design/frontend-design.md
```diff
@@ -0,0 +1,147 @@
...
+    
+    For the implementation of visulize classes: in order to print the <b>class object</b> from repl commandline, we need to add an another print function, taking the result value of the executed "repl commandline",  the result type returned by the type checking program as the paramters, and BasicRepl Class as parameters. We then use the class info and the memory object of BasicRepl Class to traverse the result value(object address) recursively and combine them into html object.
```
jpolitz: Yeah the main thing here is getting the class name of the resulting value, and then getting access to the environment/memory to extract from that. Putting all those pieces together is the work – looks like you've found the right places to look!

webstart.ts
```diff
@@ -30,6 +30,41 @@ function assert_not_none(arg: any) : any {
   return arg;
 }
 
+function get_code_example(name: string) : string {
```
jpolitz: Is there any way to store these in files and load them as part of the build process? This will be annoying to maintain. Look into webpack ways to import constant strings from files or similar

webstart.ts
```diff
@@ -41,7 +76,42 @@ function webStart() {
     ).then(bytes => 
       WebAssembly.instantiate(bytes, { js: { mem: memory } })
     );
+    
+    function console_log_class(repl:BasicREPL, pointer:number, classname:string,level:number) : Array<string>{
 
+      var fields_offset_ = repl.currentEnv.classes.get(classname);
+      var fields_type = repl.currentTypeEnv.classes.get(classname)[0];
+      var mem = new Uint32Array(memory.buffer);
```
jpolitz: I wonder if there's a way to call `load` and `store` rather than doing direct indexing here. This will be hard to maintain if memory management updates representations.

jpolitz (commented):
Great start! Looking forward to seeing more interaction w/memory plus all the other improvements.

What happens if I make a cyclic linked list and try to print it?

liusitan:
> 
printing cyclic linked list is displaying over the maximum stack size, we are considering using foldable/interactive elements to accomplish this.

webstart.ts
liusitan: sorry just found this, I understand the load part, what do you mean by the store part?

liusitan: load to get the object value out and what about store?

PR #24, opened 2022-05-07
Fancy Calling Convention design

b04505009:


design/Fancy calling conventions-design.md
```diff
@@ -0,0 +1,118 @@
...
+    parameters: Array<Parameter<A>>, 
+    ret: Type, 
+    inits: Array<VarInit<A>>, 
+    body: Array<Stmt<A>>,
+    arbarg_idx: number,
+    kwarg_idx: number,
+}
+```
+
+To deal with parameters with default value, we would like to add an attribute call `value` in `Parameter<A>` to store the default value. It could be any expression at the parsing stage. If there's no specific default value, the attribute will be set to `null`. This can also help us distinguish `None` from no default value.
```
jpolitz, line 30: Interesting! It could also be that it's a `?`-field, in which case checking if `value` is present tells us whether it's there.

design/Fancy calling conventions-design.md
```diff
@@ -0,0 +1,118 @@
...
+To deal with arbitrary arguments and keyword arguments, we would like to add attribute `arbarg_idx` to store the index of the tuple for the arbitrary arguments, and `kwarg_idx` to store the index of the dictionary for the keyword arguments. Since we will see all the arbitrary arguments as a tuple, and all the keyword arguments as a dictionary, we store the indices of them to distinguish them from normal tuple and dict parameters.
```
jpolitz, line 31: Doesn't `**kwargs` have to go last?

jpolitz: Also this allows representing `*args` and `**kwargs` with default values, which doesn't seem great.

Could we instead just have two different fields for them? Is their position relevant?

design/Fancy calling conventions-design.md
```diff
@@ -0,0 +1,118 @@
...
+The changes in IR are similar to AST, except that we could get the default value of parameters to be `Value<A>` type.
+
+## Testcases
+
+### Stage 1
+
+Support arguments with basic types and accept default value setting
+
+```python=
+def f(x : int, x : int = 5, z : int = 10): # redefine argument name
```
jpolitz: Where are the body of these functions?

jpolitz: I think this step (named args with names in assignment position) is probably enough for phase 1, **or** doing `*args` is enough for phase 1. I wouldn't try to do two of those first.

jpolitz (commented):
I want to hear more about your implementation strategy. Will you do the work in lower, how will you type-check these functions, etc? There are a number of details about that and I don't see a description of changes to the environment, etc.

I really like the idea of starting with just default values and filling those in.

If you wanted, even just _positional_ default values is a good starting point (no `x = 10` in the _caller_, just default values at the end of the arg list in the function).

Don't underestimate the type-checking details here, and that interaction with closures and stuff is upcoming.

Also, what about method parameters?

design/Fancy calling conventions-design.md
zander363: Helpful comment, we fixed our ast as your suggestion

design/Fancy calling conventions-design.md
zander363: At first, yes, the position matters.
For example, the 2 functions below would be different
```python=
def f(a, *args):
    print(a)
f(1,2,3)
```

```python=
def f(*args, a):
    print(a)
f(1,2,3)
```
Also the name of arbitrary arguments could be other name from `*args` or `**kwargs`. (maybe `*a`)
To support that, treating them as normal arguments would be more reasonable than using 2 different fields.
And for the default values part, we can do some extra check to avoid this problem.

For the position of `**kwargs`, yes, it need to be the last one.
But to check if arbitrary keyword arguments allowed and maintain the similar style with `*args`, we decide to use index to keep the information

design/Fancy calling conventions-design.md
zander363: We updated the testcases in this week progression

And our plan would be like the following:
- Stage 1 (This week)
  - default value in function definition
  - assign arguments by keywords
- Stage 2 (Future)
  - arbitrary arguments (it is dependent on tuple and dictionary, so we have to deal with this later)

PR #26, opened 2022-05-10
Add prettier

dtjong:
Run `npm run prettier-format` to format the project.

Using an editor which run prettier on save is suggested as well, vscode has an extension, vim has a plugin, etc..

I can add a .prettierrc if there are any rules you want to add.


jpolitz:
Thanks! I'm going to go ahead and merge this so we have the option.

PR #33, opened 2022-05-14
Set implementation week 7 milestones (Responded)

yiz136:


type-check.ts
```diff
@@ -69,15 +69,21 @@ export function equalType(t1: Type, t2: Type) {
...
+  );}
 
 export function isNoneOrClass(t: Type) {
   return t.tag === "none" || t.tag === "class";
 }
 
 export function isSubtype(env: GlobalTypeEnv, t1: Type, t2: Type): boolean {
-  return equalType(t1, t2) || t1.tag === "none" && t2.tag === "class" 
+  return equalType(t1, t2) || t1.tag === "none" && t2.tag === "class" || equalSet(t1,t2)
+  // set is assignable to none
```
jpolitz: I think this comment is backward (should be `None` is assignable to `"set"`)

type-check.ts
```diff
@@ -379,8 +385,77 @@ export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<n
           throw new TypeCheckError("method call on an unknown class");
         }
       } else {
-        throw new TypeCheckError("method calls require an object");
+        if (tObj.a.tag == "set"){
```
jpolitz: Would be nice to move this into a helper, and maybe have some way to add new methods without adding more `if` statements.

type-check.ts
```diff
@@ -379,8 +385,77 @@ export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<n
...
+              return {...expr,a:BOOL,obj:{...expr.obj,a:SET(NUM)}};
+          } 
+        } else{
+          throw new TypeCheckError("method calls require an object");
+        }
+      break;
       }
+      case "set_expr":
+        //Type annotation currently only takes 1 argument,
+        //Therefore we only allow a single type of variables here
```
jpolitz: Good; for sets and lists, I think that's the right decision.

set-utils.ts
```diff
@@ -0,0 +1,609 @@
...
+        "(i32.ne)", // If nodePtr not None
+        "(if",
+        "(then",
+        // There is an element, size++
+        "(local.get $size)",
+        "(i32.const 1)",
+        "(i32.add)",
+        "(local.set $size)",
+        "(local.get $nodePtr)",
+        "(i32.load)", //Loading head of linkedList
```
jpolitz, line 95: Can we use `load` and `store` from memory for this? I'm concerned about messing up the heap structure.

set-utils.ts
```diff
@@ -0,0 +1,609 @@
...
+    "(i32.const 0)",
+    "(return))",
+    "",
+    ]
+);
+
+setFunStmts.push(
+...[
+    // Basic idea is to iterate through all set entries and aggregate the number of elements followed by each entry
+    "(func $set$size (param $baseAddr i32) (result i32)",
```
jpolitz, line 43: I really don't want things like this to be concatenated strings. Can we put this in a `.wat` file like `memory.wat` is and then import it? This is really difficult to maintain/edit.

Love the idea of writing these in .wat though!

compiler.ts
```diff
@@ -203,6 +204,10 @@ function codeGenValue(val: Value<Type>, env: GlobalEnv): Array<string> {
       } else {
         return [`(global.get $${val.name})`];
       }
+    case "set":
+      return [`
+      (i32.const 10)
```
jpolitz, line 210: Why 10? I'm confused.

compiler.ts
```diff
@@ -48,8 +49,8 @@ export function compile(ast: Program<Type>, env: GlobalEnv) : CompileResult {
     funs.push(codeGenDef(f, withDefines).join("\n"));
   });
   const classes : Array<string> = ast.classes.map(cls => codeGenClass(cls, withDefines)).flat();
-  const allFuns = funs.concat(classes).join("\n\n");
-  // const stmts = ast.filter((stmt) => stmt.tag !== "fun");
+  const setFUns : Array<string> = classes.concat(setUtilFuns());
+  const allFuns = funs.concat(setFUns).join("\n\n");
```
jpolitz, line 54: I really don't want every generated program to have all of this code in it, especially not at the REPL. Let's figure out how to move this to a file.

designs/set-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+In the ir.ts, we add set to the value.
+```
+export type Value<A> = 
+    ……
+  | { a?: A, tag: "set"}
+```
+
+## Implementation and Memory Management
+The set is hashtable-like. Elements are stored using 10 linked list. Each node of the linked list consists of 4 bytes value and 4 bytes address pointing to the next node. Entries for 10 lists are consecutive on the heap. Once `x = set()` is called, 40 bytes of heap memory is assigned to x, which are 10 linked list entries.
```
jpolitz, line 39: I don't understand how this is a linked list at all – it seems like a fixed-length vector. That's a fine start, and we can get to resizing later, but it's definitely not a linked list.

Is the idea that you'll eventually add the ability to have more elements by selecting the size dynamically?

jpolitz (changes requested):
This is a good start.

I really don't want to have all this WASM text appended in each program. My main ask for you is to move these implementations into a .wat file, and use something similar to our implementation of `memory.wat` to build a `.wasm` file that can be loaded. It will be much more maintainable and scalable to do it that way.

compiler.ts
ShuchiZhang: It can be any value you would like. 10 is used to calculate the entrance of the given key.

designs/set-design.md
ShuchiZhang: Linked list is used to deal with the hash collision. The size of HashSet is dynamic by always adding an element at the end of the lists.

jpolitz: I understand this now(thanks for explaining after class!). The rest of my comments are pretty accurate, though, I think.

PR #35, opened 2022-05-14
Optimizations Week #1 : Constant folding, Constant propagations using Reaching definitions

abhilash42:
For this week, for optimizations we implemented 2 optimizations:

- Constant folding
- Constant propagation

We do this by implementing the reaching definition (in a worklist framework).

We added the test infra for optimization tests and some tests for these.

A CALLOUT: We have changed the typescript target to "es2020" as we needed it to support bigint as primitive type in TS.


optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+    return val;
+}
+
+export function evaluateBinOp(op: BinOp, leftVal: Value<any>, rightVal: Value<any>): Value<any>{
+    if([BinOp.Plus, BinOp.Minus,BinOp.IDiv,BinOp.Mul, BinOp.Gt, BinOp.Lt, BinOp.Gte, BinOp.Lte, BinOp.Mod].includes(op)){
+        if(!isTagBigInt(leftVal) || !isTagBigInt(rightVal))
+            throw new Error("Compiler Error");
+        
+        switch(op){
+            case BinOp.Plus: return {tag: "num", value: leftVal.value + rightVal.value};
```
jpolitz, line 41: bigints to the rescue! I guess that saves from checking overflow and a bunch of other annoying cases.

optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+            var optimizedValue: Value<any> = optimizeValue(stmt.cond, env);
+            return {...stmt, cond: optimizedValue};
+        case "jmp":
+            return stmt;
+        case "store":
+            return stmt;
+    }
+}
+
+//Assuming jumps if it occurs will occur at the last statement of the block
```
jpolitz, line 178: Safe assumption, thanks for writing it down.

optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+    if (env.vars.has(val.name)){
+        if (["nac", "undef"].includes(env.vars.get(val.name).tag))
+            return val;
+        val = env.vars.get(val.name).value;
+    }
+    return val;
+}
+
+export function evaluateBinOp(op: BinOp, leftVal: Value<any>, rightVal: Value<any>): Value<any>{
+    if([BinOp.Plus, BinOp.Minus,BinOp.IDiv,BinOp.Mul, BinOp.Gt, BinOp.Lt, BinOp.Gte, BinOp.Lte, BinOp.Mod].includes(op)){
```
jpolitz, line 36: This might be interesting to extend w/strings eventually!

optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+        if(!isTagBigInt(leftVal) || !isTagBigInt(rightVal))
+            throw new Error("Compiler Error");
```
jpolitz: This error is interesting; I think that w/strings we may or may not run into some issues depending on how we define values. Worth thinking a little bit about how fallthrough cases like these will work in the presence of new types.

optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+    varDefs.forEach(def => {
+        if (!dummyEnv)
+            env.vars.set(def.name, {tag: "val", value: def.value});
+        else
+            env.vars.set(def.name, {tag: "undef"});
+    });
+    return env;
+}
+
+function mergeEnvironment(a: Env, b: Env): Env{
```
jpolitz, line 239: This kind of thing is hard for me to review accurately + quickly.

The main thing I want to give as feedback is that I think you need many, many more tests, given the complexity of this code. For example, none of the tests even have >=3 variables to check if merging interesting environments could be an issue, that's just off the top of my head.

I'd like to see a bunch of tests written with the goal of _breaking_ this code.

optimization.ts
```diff
@@ -0,0 +1,433 @@
...
+    preds.set(func.body[0].label, ["VD"]);
+    succs.set("VD", [func.body[0].label]);
+    outEnvMapping.set("VD", initialEnv);
+    
+    generateEnvironments([func.body[0].label], inEnvMapping, outEnvMapping, preds, succs, blockMapping);
+
+    return [inEnvMapping, outEnvMapping];
+}
+
+export function optimizeProgram(program: Program<any>): Program<any>{
```
jpolitz, line 386: Generally the structure of this all makes sense to me, and follows patterns we set up in other places. Thanks for your attention to that!

jpolitz (changes requested):
Great start. The main thing I want to see is more tests – the optimize implementation looks reasonable but is complex, and deserves some thorough testing to make sure it's working as expected.

optimization.ts
nirmalthomas2609: True!
At the same time, towards the end if time permits, we could modify IR generation to not typecast every number as bigint depending on how tedious the changes the can get and also depending on how efficiently bignums have been implemented (if at the implementation, it behaves as an i32/i64 when it fits into the range, then I'm guessing there should be only the additional overhead at compile time).

optimization.ts
nirmalthomas2609: Sure, we will put in more tests today. Sorry about that.

Does this sound like a reasonable testing strategy?
- Generate IR using the existing code, Manually modify the same as per the expected optimizations. Assert.
If there are any other suggestions, please do let us know.

jpolitz:
I tried to do a quick resolution but I think I failed to make one update of the `Annotation` vs. `Type` distinction that the errors group added. Could you make this fix-up on your end?

ranjan-rishabh:
> I tried to do a quick resolution but I think I failed to make one update of the `Annotation` vs. `Type` distinction that the errors group added. Could you make this fix-up on your end?

We have consumed this change by the Error team. One minor call out is that we needed to add 'a' to Parameters in ast.ts because error team is adding the location metadata in 'a' (using the traverse wrapper)  but Parameters did not yet contain 'a' so it was throwing type error.

> Great start. The main thing I want to see is more tests – the optimize implementation looks reasonable but is complex, and deserves some thorough testing to make sure it's working as expected.

We have added a bunch of test cases covering different scenarios. Mainly we are doing 2 types of testing:
- First IR-IR matching which actually checks if the changes to IR from optimizations is as expected. We do this for some scenarios like constant folding, constant propagation in if-else, while and functions.
- The other type of testing checks whether after optimization the program works as expected (checked through print statements) and whether IR is same or different (we specify whether for a given test python program, optimization should change IR or not). This allows us to test some complex python programs.

jpolitz:
Thank you! This is great.

Still a few errors unresolved – the `RangeError` one might not be that interesting or might be the error group's responsibility (CC @rlisahuang), but the missing `.tag` seems concerning. See

https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/runs/6558830051?check_suite_focus=true

nirmalthomas2609:
We have fixed both the "none-is-none" and "divide-by-zero" tests to work with constant folding now. Please do let us know if any further changes need to be made.

Thanks

optimization.ts
nirmalthomas2609: Updated the error message (ideally, this should never be thrown if everything prior to that works fine - more to convince typescript of the types)

optimization.ts
nirmalthomas2609: We'll try and add support for strings and lists with constant folding by the next milestone

PR #36, opened 2022-05-14
Memory management Milestone 1

tgujar:
We were able to pass all our tests which we included last week. I noticed that some of the tests in the last week's doc actually had incorrect values for references, these are fixed in our test file in this PR. 

We planned to do most of the ref counting functions for this week with imported JS function calls. We do plan to eventually port things to WASM once we have most functionality for the GC in place and are able to support BigInt and Lists without issues, and have more extensive tests. We had to make some additions to `asserts.test.ts` to support checking the metadata we store for objects.


lower.ts
```diff
@@ -128,7 +138,7 @@ function flattenStmt(s : AST.Stmt<Type>, blocks: Array<IR.BasicBlock<Type>>, env
       var [ninits, nstmts, nval] = flattenExprToVal(s.value, env);
       if(s.obj.a.tag !== "class") { throw new Error("Compiler's cursed, go home."); }
       const classdata = env.classes.get(s.obj.a.name);
-      const offset : IR.Value<Type> = { tag: "wasmint", value: classdata.get(s.field)[0] };
+      const offset : IR.Value<Type> = { tag: "wasmint", value: classdata.get(s.field)[0] + METADATA_AMT };
```
jpolitz: Can't `METADATA_AMT` be internal to `memory.wat`? If it's consistent across objects and it's a constant, why expose it here? Just make `load` and `store` take it into account.

lower.ts
```diff
@@ -246,7 +256,7 @@ function flattenExprToExpr(e : AST.Expr<Type>, env : GlobalEnv) : [Array<IR.VarI
       }
       const className = objTyp.name;
       const checkObj : IR.Stmt<Type> = { tag: "expr", expr: { tag: "call", name: `assert_not_none`, arguments: [objval]}}
-      const callMethod : IR.Expr<Type> = { tag: "call", name: `${className}$${e.method}`, arguments: [objval, ...argvals] }
+      const callMethod : IR.Expr<Type> = {a: e.a, tag: "call", name: `${className}$${e.method}`, arguments: [objval, ...argvals] }
```
jpolitz: Nice fix, thanks.

lower.ts
```diff
@@ -261,34 +271,63 @@ function flattenExprToExpr(e : AST.Expr<Type>, env : GlobalEnv) : [Array<IR.VarI
...
+        const fieldType : IR.Value<Type> = f[1]
+        if (fieldType.tag  === "none") {
+          return 1;
+        }
+        return 0;
+      });
+
+      const fieldTypesBitString : number = generateBitString(fieldTypesArray);
+
+      let metadataStore : IR.Stmt<Type>[] = Array(METADATA_AMT)
```
jpolitz: I think this shouldn't be part of `lower`. I would rather see this be the responsibility of `alloc` to figure out. Putting it in `lower` says that each group that does a data structure thing will be responsible for calculating this...

compiler.ts
```diff
@@ -170,17 +206,24 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+        ...valStmts,
+        `(call $remove_scope)`
+      ];
 
     case "alloc":
       return [
         ...codeGenValue(expr.amount, env),
-        `call $alloc`
+        `call $alloc`,
+        `call $mem_gen_ref`
```
jpolitz: If this is the only place `mem_gen_ref` will be called, could it be moved inside `alloc`? Or does it need to be a part of this code?

designs/memory-management-design.md
```diff
@@ -0,0 +1,295 @@
...
+
+The size of the object is stored in bytes, and the number of references is stored along with the object.
+
+Update: We added an `amount` field to denote the total amount of memory allocated in bytes. `size` field holds the size of a single object in the data stored. The total number of objects associated would then be `(amount - metadatsize) / size`.
+
+### Compaction/ Defragmentation
+Performing defragmentation for the heap will move objects around in the heap. Since, we can't go about modifying all the references, each object we create will have an **immutable reference number**. We plan to maintain a mapping from this reference number to the actual location in memory. When compaction takes place we will update this mapping. All variables in the program will store the **reference number** instead of the actual address in memory.
```
jpolitz, line 277: Oh interesting. In general this is basically a strategy of boxing all variables, so that you can update the values in all the boxes. The trick in this case will be knowing when you're done with a variable (equivalently, your TODO of reclaiming reference numbers). How will you do that/

designs/memory-management-design.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Updates and Design decisions (Week 2) 
+
+We were able to pass all the tests which we had described last week. The node traversal and the scoping were the most intereresting this week and `Tests 9`, `Test 10`, `Test 4` and `Test 5` are a good representation of the cases we have accounted for. Our tests are present in the `memory.test.ts` file. We had to add a function to `asserts.test.ts` to support testing for memory management functions.
+
+Originally we has planned to do the bulk of our changes at the IR level but we ended up implementing code to add metadata by constructing IR objects. We found that doing so reduced the compexity of our implementation.
```
jpolitz: I'd like to hear more about why – it seems like adding the metadata in `compiler.ts` would be basically the same work (type to bitfield), and just involve modifying `alloc`.

tests/memory.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    class Rat(object):
+        id: int = 123
+        y: int = 0
+        def __init__(self: Rat):
+            self.y = 1
+
+    x: Rat = None
+    x = Rat()
+  `, [
+    // first value in the tuple denotes id, NOTE: this is a hack since we dont have access to object names
```
jpolitz, line 17: I'm a little confused about this – if you have heap access isn't it deterministic where the first-created object will go on the heap, so you can just check heap contents?

tests/memory.test.ts
```diff
@@ -0,0 +1,208 @@
...
+            self.y = 1
+    x: Rat = None
+    y: Rat = None
+    z: Rat = None
+    x = Rat()
+    y = x
+    z = y
+    `, [
+    // first value in the tuple denotes id, NOTE: this is a hack since we dont have access to object names
+    [123, refNumOffset, 3], // 1 reference in the program where object id is 123
```
jpolitz: Should this say "3 references in the program where...". Copy-paste is confusing to read.

tests/memory.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    x: Link = None
+    y: Link = None
+    x = Link()
+    y = x.add(456)
+    `, [
+    // first value in the tuple denotes id, NOTE: this is a hack since we dont have access to object names
+    [123, refNumOffset, 1], // 1 reference in the program where object id is 123
+    [456, refNumOffset, 2], // 2 reference in the program where object id is 456
+    [123, typeOffset, 2]
```
jpolitz: So this 2 is saying the 2nd field is a reference (`0x00000002` in metadata), right?

tests/memory.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    x: Link = None
+    x = Link()
+    x.add()
+    x = None
+    `, [
+    // first value in the tuple denotes id, NOTE: this is a hack since we dont have access to object names
+    [123, refNumOffset, 0], // 0 references in the program where object id is 123
+    [456, refNumOffset, 0], // 0 references in the program where object id is 456
+    ]); // all types are values or non-references
```
jpolitz: This comment is weird. There are definitely references, right?

jpolitz (changes requested):
This is generally great. I don't want to demand many changes, but my feedback is twofold:

1. I think reclaiming references numbers to do good defragmentation is going to be as hard as tracking the stack. Maybe when you do the pop_stack thing you can do something with all the new variables lots made since the last push_stack? Might work.
2. I really don't like the calculation of the bitfield in `lower`. I feel like that should be more hidden behind the ir/memory.wat interface. Can you say more about why putting it in lower was easier? Is there a change to the IR that would make it easier to put it in the compiler rather than lower?

tests/memory.test.ts
tgujar: whoops sorry about that, I'll fix this

tests/memory.test.ts
tgujar: yes

tests/memory.test.ts
tgujar: We decrement the reference count when we set `x = None` so there would be no way to access those created objects

tests/memory.test.ts
tgujar: When we do start reclaiming memory and move stuff around in the heap, I thought `id` would be easier to write tests with. We could totally switch to calculating addresses though, but I feel it makes writing more complex tests a bit harder. For e.g if we move stuff around in the heap, we would have to know where it got moved to so that we can check the values at that address

lower.ts
tgujar: yep, will fix this

compiler.ts
tgujar: we can move it inside alloc

designs/memory-management-design.md
tgujar: We know when we are done with a variable when we run `traverseUpdate` and the reference count or the value in the first 4 bytes of the object hit zero. To reclaim ref numbers I thought I could just keep a `Set()` of reference numbers which are available to be reused and assign new ref numbers from there if its not empty

tgujar:
I think we can move stuff to the ir side instead of ast. We thought it easier since we already had `alloc` in memory.wat so we could get away with adding IR calls. But we totally see how this would cause problems for other groups and we are making the changes to move it "post IR".

As for reference numbers, the case we had in mind was something like
```
x1: Rat = None
x2: Rat = None
x3: Rat = None
x1 = Rat()
x2 = Rat()
x3 = Rat()
x2 = None
```
In this case we could retake the memory space allocated for the `x2 = Rat()` object, but we would have to shift the object `x3 = Rat()` to a different memory and we can't do that since we don't have access to what value gets stored in `x3` after it has been allocated. I am not sure if this is an overkill for just this one case.

Currently we do decrease the reference count of a variable once it goes out of scope, and we could reclaim the reference numbers associated with them by just decreasing the global reference count. In this case the reference numbers aren't really necessary. But I am unsure how I could handle cases like above when I can't *walk* the stack.

tests/memory.test.ts
tgujar: I'll change the comments to say `at the end of the program`

PR #37, opened 2022-05-14
I/O milestone1 implementation

Chao-Te:
We finish the first version of I/O


IO_File/FileParser.ts
```diff
@@ -0,0 +1,121 @@
...
+        name:"__init__",
+        parameters:[{name:"self", type:{tag:"class", name:"File"}}, 
+                    {name:"fd", type:NUM}],
+        ret: NONE,
+        inits:[],
+        body:[{tag:"field-assign", obj:{tag:"id", name:"self"}, field:"fd", value:{tag:"id", name:"fd"}}], // self.fd = fd
+    }
+}
+*/
+function getFileInit(): FunDef<null> {
```
jpolitz, line 58: This is interesting! Could almost implement this as a Python library and read it in/parse it with our parser, right? It feels like there should be some way to get this to work more cleanly than having to write out ASTs by hand.

jpolitz: Like I almost wish we could leave out this infrastructure and just provide it through the built-in functions at first, then separately layer on a uniform strategy for these.

IO_File/FileSystem.ts
```diff
@@ -0,0 +1,154 @@
...
+let fdCounter = 0;
+let fs = new Map<number, OpenFile>(); // track current open files
+
+/*
+ * TODO Later: The input to open should be some value for string
+ * right now pretend that we are creating a file without a fname string
+ * 
+ * ct: for the mode, we might need to create a 'mode translator' to translate from Python mode to our mode
+ */
+export function open(filePathAddr: number, mode: number): number {
```
jpolitz, line 44: Cool, like the simple use of numbers to simulate paths.

IO_File/FileSystem.ts
```diff
@@ -0,0 +1,154 @@
...
+        currentPosition: 0,
+        mode: mode, // random mode
+        fileSize: 0, // according to the current test case we should assign 0
+    });
+
+    return fdCounter - 1;
+}
+
+export function read(fd: number, numByte: number): number {
+    console.log(`read is called, fd: ${fd}, numByte: ${numByte}`);
```
jpolitz: Stray console.log to remove

IO_File/FileSystem.ts
```diff
@@ -0,0 +1,154 @@
...
+
+enum FileMode {
+    OPEN,   // Just open the file
+    R_ONLY, // read only mode
+    W_APPEND,  // append the data to the end
+    RW,     // read and write mode
+    W_CURR, // write to the current position. If wwe have data in the current position, we overwrite that piece of data  
+}
+
+export const buildin_file_libs = `
```
jpolitz, line 22: Interesting (and pretty useful) to specify this here. Neat.

ast.ts
```diff
@@ -1,5 +1,7 @@
 // import { TypeCheckError } from "./type-check";
 
+import { createModuleResolutionCache } from "typescript"
```
jpolitz: Accidental import?

run_remove_js.sh
```diff
@@ -0,0 +1,16 @@
+# ./node_modules/.bin/tsc
+# node debugstart.js
+# remove the create .js file
```
jpolitz: Why is this needed? If it is, we should add it as `make clean`

webstart.ts
```diff
@@ -51,14 +52,20 @@ function webStart() {
...
+        buildin_close: close,
+        buildin_seek:  seek,
       },
       libmemory: memoryModule.instance.exports,
       memory_values: memory,
       js: {memory: memory}
     };
     var repl = new BasicREPL(importObject);
-
+    console.log("importObject");
```
jpolitz: stray, removable logs

designs/io-file.design.md
```diff
@@ -0,0 +1,49 @@
...
+
+  def close(self:File):
+
+open(filePath: int, mode:int)
+```
+
+
+Since the file system is complicated, we put most of our file system code in JavaScript and provide the api for WASM code to call the function. We choose to implement most of the file system in JavaScript because it will make the implementation simpler and easier to debug. For the File class, which represents the File class in Python, for the user to use. We create the parsed result for the function and class and import them into the parser via a function call. This can make sure that we make minimum modifiction to the major code.
+
+As we do not have the support of string, we use some dummy data and file to test our code. Namely, every single time when we open() a file, we create a dummy txt file called "test.txt" and do read() and write() to it later. The file will be actually stored in the FileSystem API in browsers. This also will make us cannot write normal unit test to show whether the test case pass or not. Instead, we will show our test cases passed by showing the context in window.localStorage in our browser in screenshots.
```
jpolitz, line 25: I think we could still have something on startup that has mappings from numbers to files with actual contents, and then read that contents, right? That would be useful infrastructure to get in place. (What's in this PR is a great start anyway, but I think we could manage that much).

jpolitz (changes requested):
This is pretty good, nice work! There are some cleanups and questions I have throughout to address/follow up on.

IO_File/FileParser.ts
Chao-Te: Sure! We will do that!

ast.ts
Chao-Te: We will check it out. I think it is added accidentally.

run_remove_js.sh
Chao-Te: Yes, I think we might need to have make clean. We meet some problem when trying to fix the testing code issue.

designs/io-file.design.md
Chao-Te: Got it.

PR #38, opened 2022-05-14
Testing-implementation

jasoonn:


designs/testing-design.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+#### Test Case 4  (Classes, Strings , Builtin Fncs)
+```
+class Start:
+  a : string = ""
+  b : string = ""
+  def set(self, a : string , b : string):
+     self.a = a
+	 self.b = b
```
jpolitz: Is this an indentation error?

designs/testing-design.md
```diff
@@ -0,0 +1,251 @@
...
+```
+False
+True
+```
+
+## Changes we want to make
+#### changes in tests/
+- Add more integrated test cases
+- Add tests on different os(macOS and windows)
+- Add automation testing for webApp(Use Selenium or other tools)
```
jpolitz, line 251: Selenium would be awesome! I think that's a great next milestone.

tests/integreated.test.ts
```diff
@@ -0,0 +1,330 @@
+// import { assertPrint, assertFail, assertTCFail, assertTC } from "./asserts.test";
```
jpolitz, line 1: This is great. I think including these and slowly uncommenting them as they work is a terrific idea.

jpolitz (commented):
This is great, and will be a terrific resource as we build up more.

I think getting something running on Selenium would be a great next step!

designs/testing-design.md
jasoonn: Yes. Thank you for pointing that out. I have fixed the indentation in the latest commit.

designs/testing-design.md
jasoonn: Got it! We will put them in the next milestone.

PR #39, opened 2022-05-14
Bignums milestone1

taipeiyuka:


compiler.ts
```diff
@@ -138,13 +139,30 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
     case "binop":
       const lhsStmts = codeGenValue(expr.left, env);
       const rhsStmts = codeGenValue(expr.right, env);
-      return [...lhsStmts, ...rhsStmts, codeGenBinOp(expr.op)]
+      if (expr.op == BinOp.Is || expr.op == BinOp.And || expr.op == BinOp.Or) {
+        return [...lhsStmts, ...rhsStmts, codeGenBinOp(expr.op)]
+      }
+      // return [...lhsStmts, ...rhsStmts, codeGenBinOp(expr.op)]
+      var callName = codeGenBinOp(expr.op);
```
jpolitz: I'm confused – why not have `codeGenBinOp` return the call instruction itself? Why have the condition above that checks the various op types at all?

compiler.ts
```diff
@@ -138,13 +139,30 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+      return [...lhsStmts, ...rhsStmts, `(call $${callName})`]
 
     case "uniop":
       const exprStmts = codeGenValue(expr.expr, env);
       switch(expr.op){
         case UniOp.Neg:
-          return [`(i32.const 0)`, ...exprStmts, `(i32.sub)`];
+          return [
```
jpolitz, line 149: Why are the ops above function calls, but this is inline WASM?

compiler.ts
```diff
@@ -190,7 +208,42 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+      var i = 0
+      var return_val : string[] = []
+      
+      return_val.push(`(i32.const ${n})`);
+      return_val.push(`(call $alloc)`);
+      return_val.push(`(local.set $$scratch)`);
+      
+      // store the bignum in (n+1) blocks
+      for (i; i < n; i++) {
+        if (i == 0) {
```
jpolitz: This is awkward. Why put this special case inside the loop? Could just start `i` at 1 for the loop, for example. Also, if you want it in the loop to share the get/const/store lines, those shouldn't be repeated.

compiler.ts
```diff
@@ -209,27 +262,27 @@ function codeGenValue(val: Value<Type>, env: GlobalEnv): Array<string> {
 function codeGenBinOp(op : BinOp) : string {
   switch(op) {
     case BinOp.Plus:
-      return "(i32.add)"
+      return "add"
```
jpolitz: Why not make these `(call $add)`, `(call $mul)`, etc?

lower.ts
```diff
@@ -73,7 +73,7 @@ function lowerClass(cls: AST.Class<Type>, env : GlobalEnv) : IR.Class<Type> {
 function literalToVal(lit: AST.Literal) : IR.Value<Type> {
     switch(lit.tag) {
         case "num":
-            return { ...lit, value: BigInt(lit.value) }
+            return { ...lit, value: BigInt((lit.value).toLocaleString('fullwide', {useGrouping:false})) }
```
jpolitz, line 77: I don't understand this line – could you add a comment or explain in more detail why it's needed? What's a test that would fail without this line and why?

tests/bignum-test.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    test_print(BigInt("100000000000000000000000"));
+    test_init_assign(BigInt(0), BigInt(10), BigInt("100000000000000000000000"));
+    test_binop("3_test_greater_than", BigInt("100000000000000000000000"), BigInt("100000000000000000000001"), ">", "True");
+    test_binop("4_test_less_than", BigInt("100000000000000000000000"), BigInt("100000000000000000000001"), "<", "False");
+    test_binop("5_test_greater_equal", BigInt("100000000000000000000000"), BigInt("100000000000000000000001"), ">=", "True");
+    test_binop("6_test_less_equal", BigInt("100000000000000000000000"), BigInt("100000000000000000000001"), "<=", "False");
+    test_binop("7_test_equal", BigInt("100000000000000000000000"), BigInt("100000000000000000000000"), "==", "True");
+    test_binop("8_test_not_equal", BigInt("100000000000000000000000"), BigInt("100000000000000000000001"), "!=", "True");
+    test_uniop("9_test_negate", BigInt("100000000000000000000000"), "-", "-100000000000000000000000");
+    test_param_return("10_test_param_return", BigInt("100000000000000000000000"), BigInt(0), "100000000000000000000000");
```
jpolitz: In your design review, I gave some specific test cases to add. These don't include any cases where you cross from a small to a big number, or test things around the boundary at `2^32`. Where are those?

tests/bignum-test.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        `x : int = ${x}
+        print(${op}x)`, 
+        [`${res}`]);
+}
+
+export function test_param_return(des: string, c: bigint, param: bigint, res: string) {
+    assertPrint(des, 
+        `def f(c:int) -> int:
+            c = ${c}
+            return c
```
jpolitz, line 42: Thank you for making tests in functions. It would be useful to make sure the _other_ operations that were tested at the top level also work in functions. Since you've done the work of the helpers here (which is nice), maybe there's a way to test this even more by plugging the same tests into different contexts?

tests/bignum-test.test.ts
jpolitz: Also, these are _very_ specific numbers. What about some even-bigger numbers? What about numbers with digits other than 0 and 1? What about negative numbers turning into positive numbers? What about tests for add, subtract, multiply, and divide?

tests/import-object.test.ts
```diff
@@ -42,14 +46,24 @@ export const importObject : any = {
...
-    print_bool: (arg: number) => print(Type.Bool, arg),
-    print_none: (arg: number) => print(Type.None, arg),
-    abs: Math.abs,
-    min: Math.min,
-    max: Math.max,
-    pow: Math.pow,
+    print_num: (arg: number) => print(Type.Num, arg, importObject.libmemory.load),
+        print_bool: (arg: number) => print(Type.Bool, arg, null),
+        print_none: (arg: number) => print(Type.None, arg, null),
+        abs:  (arg: number) => builtin_bignum([arg], bigMath.abs, importObject.libmemory),
```
jpolitz: The format below is nice – good use of `libmemory`, helpers to have clean code for these.

tests/project-test.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    test_print(BigInt(100000000000000000000000));
+    test_init_assign(BigInt(0), BigInt(10), BigInt(100000000000000000000000));
+    test_binop("3_test_greater_than", BigInt(100000000000000000000000), BigInt(100000000000000000000001), ">", "True");
+    test_binop("4_test_less_than", BigInt(100000000000000000000000), BigInt(100000000000000000000001), "<", "False");
+    test_binop("5_test_greater_equal", BigInt(100000000000000000000000), BigInt(100000000000000000000001), ">=", "True");
+    test_binop("6_test_less_equal", BigInt(100000000000000000000000), BigInt(100000000000000000000001), "<=", "False");
+    test_binop("7_test_equal", BigInt(100000000000000000000000), BigInt(100000000000000000000000), "==", "True");
+    test_binop("8_test_not_equal", BigInt(100000000000000000000000), BigInt(100000000000000000000001), "!=", "True");
+    test_uniop("9_test_negate", BigInt(10000000000000000000000), "-", "-10000000000000000000000");
+    test_param_return("10_test_param_return", BigInt(10000000000000000000000), BigInt(0), "10000000000000000000000");
```
jpolitz: Is this just a duplicate file of tests?

utils.ts
```diff
@@ -1,5 +1,152 @@
...
+    return value1 > value2
+  },
+}
+
+export function binop_bignum(args: number[], builtin: Function, libmem: WebAssembly.Exports): number {
+  var rslt : bigint = BigInt(0);
+  const load = libmem.load;
+  // if(args.length === 1)
+  //   rslt = builtin(load_bignum(args[0], load));
+  // else
```
jpolitz: Stray comment to clean up.

utils.ts
```diff
@@ -1,5 +1,152 @@
...
+  const load = libmem.load;
+  // if(args.length === 1)
+  //   rslt = builtin(load_bignum(args[0], load));
+  // else 
+  
+  if(args.length === 2)
+    rslt = builtin(load_bignum(args[0], load), load_bignum(args[1], load));
+  else
+    throw new Error("Runtime Error: too many arguments for builtin functions");
+  console.log("binop comp result:", rslt)
```
jpolitz: Stray console log to clean up

utils.ts
```diff
@@ -1,5 +1,152 @@
...
+  },
+  lt(value1: bigint, value2: bigint) {
+    return value1 < value2
+  },
+  gt(value1: bigint, value2: bigint) {
+    return value1 > value2
+  },
+}
+
+export function binop_bignum(args: number[], builtin: Function, libmem: WebAssembly.Exports): number {
```
jpolitz, line 60: This library is pretty well-designed. Good use of `libmem` and helpers to create + allocate new numbers.

compiler.ts
```diff
@@ -190,7 +208,42 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
 function codeGenValue(val: Value<Type>, env: GlobalEnv): Array<string> {
   switch (val.tag) {
     case "num":
-      return ["(i32.const " + val.value + ")"];
```
jpolitz, line 192: Would it be possible to implement this all in `lower`, using `alloc` and `store` and `wasmint`? What are the tradeoffs with doing that?

compiler.ts
jpolitz: Also, could you implement this in `lower`, by just directly compiling to the correct `call`?

compiler.ts
jpolitz: Also, what if a user defines a function called `mul`?

jpolitz (changes requested):
Overall, this design is solid. I get the idea of boxing all the numbers and going to and from the JS representation.

This is probably not fast, which is fine for this milestone :-)

It's worth talking to the memory management group and the lists/strings groups about how numbers will be represented and work as offsets when they are bignums. It might be that you collaborate a little bit to provide a helper in that case.

I think the way the binop functions are called needs some attention and could be simplified.

I also think the testing is really incomplete (I don't even see addition tested). Please add much more thorough tests (there should be at least dozens). I really like the helpers you wrote for the tests – use them to make it easy to write more!

lower.ts
taipeiyuka: There is an unexpected error when passing the number directly to BigInt.
```
> BigInt(1e23)
99999999999999991611392n
> BigInt(100000000000000000000000)
99999999999999991611392n
> BigInt("100000000000000000000000")
100000000000000000000000n
```

compiler.ts
taipeiyuka: The binops are function calls to Javascript. To negate a bignum, we invert the sign of the metadata (length). This can be done with simple WASM code.

compiler.ts
taipeiyuka: The current design allows constant folding in optimization. If bignums were lowered into multiple `wasmint`'s and `store`'s, it would create extra cost to reconstruct the constant value. I believe it would be a lot harder to maintain.

The benefit of implementing this in `lower` is that binop functions can be optimized. This can be approximated by writing good quality binop functions in `compiler.ts`.

Other than that, we are discussing with the memory management group to support our calls to `memory.wat`.

compiler.ts
AmberWow123: This part is now fixed. The special case when i = 0 in the for loop is now moved to the place before the for loop, and the for loop is now looping from i = 1 to i = n-1.

tests/project-test.test.ts
AmberWow123: Yes, it is now removed. Sorry for the confusion.

compiler.ts
AmberWow123: Sure, thanks for the recommendations! For the binary operations that interact with bignums in codeGenBinOp function, it is now modified to return in the format of `call $$${binop_name}`. For example, for add operation, it will be like `call $$add`. Obviously, we modified the function name of binary operation from `add` to `$add` because it can thus be distinguishable from a function call `add`.

compiler.ts
AmberWow123: For the first question, having such a condition checking if ops is `eq`, `and`, or `or` is because we only defined the binary operations that interact with bignums. For those that don't interact with bignums, we just keep it the same. Therefore, for ops like `eq`, `and`, and `or`, it stays the same; but, for other binary operations, we then call the corresponding functions to handle the operations with bignums.

taipeiyuka: We cleaned the code up a little so that everything is wrapped in `codeGenBinOp` now.

tests/bignum-test.test.ts
taipeiyuka: We added a couple of test cases including the ones you mentioned in our design proposal. All binary operations are tested wrapped in function declarations and works with negative values.

tests/bignum-test.test.ts
AmberWow123: We have now added more testcases.

PR #40, opened 2022-05-14
String - Milestone 1 Week7

williammm001:


type-check.ts
```diff
@@ -243,12 +257,38 @@ export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<n
         case BinOp.Mul:
         case BinOp.IDiv:
         case BinOp.Mod:
+          if(equalType(tLeft.a, STRING) && equalType(tRight.a, STRING)){
+            if (tLeft.tag == "literal" && tLeft.value.tag == "str" && tRight.tag == "literal" && tRight.value.tag == "str"){
+              console.log("+ with STR");
```
jpolitz: Stray `console.log`

webstart.ts
```diff
@@ -18,10 +21,25 @@ function stringify(typ: Type, arg: any) : string {
 
 function print(typ: Type, arg : number) : any {
   console.log("Logging from WASM: ", arg);
-  const elt = document.createElement("pre");
+  if (typ.tag == "str") {
```
jpolitz, line 25: This condition might be good to discuss with the frontend group – I'm not totally sure I understand why this is needed (a case where a comment could help a lot). I'm assuming this and `print_enter` go together, but I don't understand why.

webstart.ts
```diff
@@ -1,14 +1,17 @@
...
 
 function stringify(typ: Type, arg: any) : string {
   switch(typ.tag) {
     case "number":
       return (arg as number).toString();
     case "bool":
       return (arg as boolean)? "True" : "False";
+    // convert from ASCII to Character
+    case "str":
+      return String.fromCharCode(arg as number);
```
jpolitz, line 15: Doesn't this just convert a _single_ character, not a whole string?

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+    ```
+    Expected output:
+    ```Python
+    abcd
+    "abcd"
+    ```
+
+    **Update Week 7**:
+
+    We are considering that using a series of ASCIIs to store string in the momeory. However, in ir.ts, type VarInit\<A> = { a?: A, name: string, type: Type, value: Value\<A> }, the value is of type Value\<A>. Thus, for now we cannot simply assign the string to the variable (s here) because we need to allocate a block of memory and store the ASCIIs (this should be done in Expr\<A>).
```
jpolitz, line 18: This is a reasonable restriction!

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+    1. When initializing a string variable (s:str = "abcd"), in parser.ts, we first parse it to s:str = None, then s = "abcd", i.e. we spilt varInit into varInit and Stmt(assignStmt). However, in class, we just have varInits and methods, thus, we cannot modify the string initialization in this way.
+    2. We are thinking about changing the IR.VarInit\<A> = { a?: A, name: string, type: Type, value: Value\<A> } to IR.VarInit\<A> = { a?: A, name: string, type: Type, value: Expr\<A> }. By this way, we can accomplish the allocation and storing jobs.
+    3. Actually, we are thinking about also change the AST.VarInit\<A> = { a?: A, name: string, type: Type, value: Literal } to AST.VarInit\<A> = { a?: A, name: string, type: Type, value: Expr\<A> }. We know that in ChocoPy, when initialzing the variables, we should not use a expression at the right hand side. Thus, when initialzing the variables which need the memory, the right hand side can only be None. However, by modifying the VarInit stucture, we can support the initializations such as l: [int] = [1, 2, 3] and c: C = C() where C is a class object. Then our behavior will be more python-like.
```
jpolitz, line 23: Why might this restriction exist? What other groups might it affect? Are there any expressions that are particularly tricky to put on the right-hand side?

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+    4
+    ```
+
+    **Update Week 7**:
+
+    Our solution is that when storing a string, we firstly store its length on top of its memory block (The string type should be immutable, so the length will not change). By doing so, when we want to load the string, we can know where to stop when generating the wasm loop block.
+
+    Besides, regrading the builtin print and len, we would like to define two functions in wasm: print_str and len_str to handle the string case. For print_str, we first get the length of the string, and then generate a wasm loop to load the ASCIIs of the strings and convert them back to string. For len_str, simpler, we can get the length and call print_num (haven't implemented).
+
+    For now, we implement the print_str in complier.ts by adding new variables to get the length of the string and handle the loop block in wasm. In the following weeks, we should implement them in lower.ts using
```
jpolitz, line 56: I think it's fine for `print` to stay as an external function, if you want my opinion!

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+    print(s1 + s2)
+    ```
+    Expected output:
+    ```Python
+    abcd
+    ```
+
+    **Update Week 7**:
+
+    One main issue of the operations (concat, comparsion, indexing) on string is that we only have the information of the string starting address. If we want to realize these operations without touching the complier.ts, I think we need more information of string.
```
jpolitz, line 82: Did implementing `print` require modifying `compiler.ts`? Could you do something similar (use functions) for append/comparison/indexing?

For indexing why not use `load`?

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+
+    We have some solutions as followed:
+    1. Modify some structures in AST of IR by adding the string value information, but it seems problematic because it will change some basic behaviors of our complier.
+    2. Adding Map\<string, string>() in the envirnment in type-check.ts and lower.ts. When we encounter a string variable, we not only set its varaible map to true, but also set its value map to its value. Because we need the value information rather than the starting address to do the operations above.
```
jpolitz, line 86: Consider this program:

```
def f(x : str, y : str) -> str:
   return x + y
f("a", "b")
f("abc", "def")
```

What would be in that map for `x` and `y`?

It might be useful to refer back to lecture on Tuesday, May 10 for some discussion on making sure to test your features within functions.

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+
+    **Update Week 7**:
+
+    We will look into the escape sequences next week.
+
+## Changes on AST, IR and builtin libraries
+### AST
+- We need to add a new Type "string" on the ast. It should be "{tag: "string"}".
+- We need to add a new Literal "string". It should be "{a?:A, tag: "string", value: string}", where a is the annotation for type checking, the value is the string itself. For example, the string "abc"'s literal value is "abc".
+- We need to add a new expression for indexing "{a?: A, tag: "indexing", name: Expr\<A>, index: Expr\<A>}". For example, if we enter "s[0]", the name will be an id expression on the "s" case, the index will be 0. The reason that we need the "index" to be an expression is that we also allow the script like "s[0+1]" to be acceptable.
```
jpolitz, line 206: The current ast.ts has this already: https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/40/files#diff-08420cd45e72258159c0a8701934f95b0e70129064a4867914f3a3b5782a0e5fL40

designs/Strings-design.md
```diff
@@ -0,0 +1,265 @@
...
+
+
+### IR
+- We need to add a new Value type which is like "{ tag: "string"; value: string; address: number }". The value is the string itself as in ast. The address is where the start of the addresss where the string is stored.
```
jpolitz, line 210: See the function example above.

parser.ts
```diff
@@ -27,6 +34,10 @@ export function traverseLiteral(c : TreeCursor, s : string) : Literal {
 
 export function traverseExpr(c : TreeCursor, s : string) : Expr<null> {
   switch(c.type.name) {
+    case "String":
+      // const str_ = s.substring(c.from,c.to).replace(/["']/g, "")
```
jpolitz: These comments should probably be removed.

runner.ts
```diff
@@ -117,11 +121,14 @@ export async function run(source : string, config: Config) : Promise<[Value, Glo
     ${config.functions}
     ${compiled.functions}
     (func (export "exported_func") ${returnType}
+      (local $i i32)
+      (local $length i32)
```
jpolitz: I'm confused about these hardcoded variables – what are they for? Wouldn't they also be needed in every function definition? Why can't new temporaries introduced in `lower.ts` do the work of these?

tests/pa3-visible.test.ts
```diff
@@ -33,7 +33,6 @@ print(x)` , [`5`]);
     `
 class C(object):
   x : int = 123
-
```
jpolitz: I think it would be better not to introduce diffs into this test file. (Was something broken that needed these newlines removed?)

type-check.ts
```diff
@@ -243,12 +257,38 @@ export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<n
...
+            }
+          }
           if(equalType(tLeft.a, NUM) && equalType(tRight.a, NUM)) { return {a: NUM, ...tBin}}
           else { throw new TypeCheckError("Type mismatch for numeric op" + expr.op); }
         case BinOp.Eq:
         case BinOp.Neq:
           if(tLeft.a.tag === "class" || tRight.a.tag === "class") throw new TypeCheckError("cannot apply operator '==' on class types")
-          if(equalType(tLeft.a, tRight.a)) { return {a: BOOL, ...tBin} ; }
+          if(equalType(tLeft.a, tRight.a)) {
+            console.log("BOP",tBin);
```
jpolitz: Stray `console.log`

jpolitz (changes requested):

jpolitz:
Good start. There are some things that need to be resolved before merging:

- Make sure you are using the common `index` AST form for all indexing expressions (the one that has `Expr` in value position) that we introduced.
- Make sure your _design_ for strings will work within functions
- Consider using functions for comparison, concatenation, etc.
- Address other specific code comments

designs/Strings-design.md
williammm001: Yes we have used the "index" expr already.

designs/Strings-design.md
williammm001: Now we have solved the problem now. A string can be defined in this way now in body, functions and classes: a:str = "abcd".

designs/Strings-design.md
williammm001: Yes now it is a external function.

designs/Strings-design.md
williammm001: Now we have fixed the problem. We add variants of allocate and store, now they can receive an expression as the value. The way we do this is because we don't know the length of a string in this case: **a = b[1]**. (Both a and b are strings). So, we have to use the **load** to get the value. Now, concat, comparison and index are implemented and tested.

designs/Strings-design.md
williammm001: image

Now all the features have been tested within functions and classes.

designs/Strings-design.md
williammm001: Solved.

runner.ts
williammm001: We have now remove these two lines. It is used for the **print_str**. Now the new print is encapsulated as an external function.

tests/pa3-visible.test.ts
williammm001: We mistakenly modified this file.Fixed now.

webstart.ts
AnthonyHoo: We changed the way to print a string. At the beginning of printing a string, we will input ascii code 256 as the symble. Then the print function will create a "pre" label and store this element in <output> lable. Then, every time we load a ascii code from the memory, this function will convert the ascii code to a character. Get the last child of "output", which is the place where we store the information of string at front end. In this way, we can wrap a string with only one "pre" lable and we don't need print_enter to change the line anymore, because this label will take such effect.

webstart.ts
AnthonyHoo: That's true, because we call this method every time we load a ascii code from the memory. The result will be returned and recorded in the same  "pre" lable. Then we can show the whole string.

ir.ts
```diff
@@ -18,22 +18,25 @@ export type Stmt<A> =
   | {  a?: A, tag: "pass" }
   | {  a?: A, tag: "ifjmp", cond: Value<A>, thn: string, els: string }
   | {  a?: A, tag: "jmp", lbl: string }
-
   | { a?: A, tag: "store", start: Value<A>, offset: Value<A>, value: Value<A> } // start should be an id
-
+  | { a?: A, tag: "store_str", start: Value<A>, offset: Value<A>, value: Expr<A> } // start should be an id
```
jpolitz: Why can't the value be an identifier? The idea of lower is that you'd introduce any intermediate variables you need so that you can use `Value` in these positions.

ir.ts
```diff
@@ -18,22 +18,25 @@ export type Stmt<A> =
...
+  | { a?: A, tag: "duplicate_str", source: Value<A>, dest: Value<A>}//duplicate the source string into the destination address
+  
 export type Expr<A> =
   | {  a?: A, tag: "value", value: Value<A> }
   | {  a?: A, tag: "binop", op: BinOp, left: Value<A>, right: Value<A>}
+  | {  a?: A, tag: "str_compare", op: Value<A>, left:Value<A>, right: Value<A> } // op: 0 for inequal and 1 for equal
```
jpolitz, line 26: Why can't these all be implemented as built-in functions?

ir.ts
```diff
@@ -18,22 +18,25 @@ export type Stmt<A> =
...
   | {  a?: A, tag: "uniop", op: UniOp, expr: Value<A> }
   | {  a?: A, tag: "builtin1", name: string, arg: Value<A> }
   | {  a?: A, tag: "builtin2", name: string, left: Value<A>, right: Value<A>}
-  | {  a?: A, tag: "call", name: string, arguments: Array<Value<A>> } 
-
+  | {  a?: A, tag: "call", name: string, arguments: Array<Value<A>> }
   | {  a?: A, tag: "alloc", amount: Value<A> }
   | {  a?: A, tag: "load", start: Value<A>, offset: Value<A> }
-
+  | {  a?: A, tag: "alloc_expr", amount: Expr<A> }
```
jpolitz: Why can't the amount to allocate be stored in an identifier?

ir.ts
```diff
@@ -18,22 +18,25 @@ export type Stmt<A> =
...
+  | {  a?: A, tag: "getLength", addr1: Value<A>, addr2:Value<A>}
```
jpolitz: Similarly, why can't this be a built-in function?

jpolitz:
Thanks for all the updates! I dropped a few more comments in – I don't think the IR changes are needed.

ir.ts
williammm001: The reason is that the "store_str" is used in this situation:
```
x:str = "abcd"
c:str = ""
x = "xyz"
c = x[1]
```
In this case, in lower.ts we cannot get the value of ```x[1]``` directly. Therefore, we need to load it from the memory. So, we use the ```store_str``` expression which can receive an ```load``` expression as the value. In this way, we can use the same ```store``` function defined in memory.wat. 

I agree that the ```value``` of ```store_str``` can be a ```Value```. In the way, we still need the ```store_str``` and we need to write a new```store_str``` wat code which similar to the defined ```store``` but it will encapsulate the ```load``` expression in the wat level. 

I think both of the solution can achieve the goals. The reason why we choose the current solution is that it can reuse the ```store``` wat code without writing a new one.

williammm001: Oh I think I have got your point. We can create a new variable and use the ```assign``` stmt (which can receive ```Expr``` as its value) to assign the ascii code to it. Then we can use the original ```store``` in both ```ir.ts``` and ```memory.wat``` without further changes.

williammm001: Fixed. ```store_str``` is removed.

ir.ts
williammm001: Fixed. ```alloc_expr``` is removed.

ir.ts
williammm001: Actually getLength is misnamed. It is not the ```len()``` function but a external wat function that can load the lengths of two strings and return their sum. ```len()``` is implemented as a builtin function.

ir.ts
Yongxing-Chen: Yeah. Actually, it is a built-in function implemented using wasm, we can do some reformatting later.

jpolitz: I'm trying to merge this, and can't really figure out what to do. It looks like `str_compare` got moved to a new IR form... but I don't understand that either. The IR form has new stuff in the compiler that refers to a built-in `str_comparison` (great), but then does extra comparisons and has another special `op` field.

Why can't the design be that string comparisons are just a kind of binop that lowers to a call to a WASM function like `str_comparison` directly? Then we don't need an expr _or_ and IR form for it.

parser.ts
```diff
@@ -79,6 +80,35 @@ function wrap_locs<T extends Function>(traverser: T, storeSrc: boolean = false):
...
+            charArr.push(str_[i-1] + str_[i])
+            console.log("push")
+            console.log(charArr[charArr.length-1])
+          }
+        }
+      }
+      return{
+        tag: "str",
+        value: charArr,
+        length: charArr.length
```
jpolitz, line 109: Does `charArr` ever change length once created? Why store the length separately?

stdlib/memory.wat
```diff
@@ -1,22 +1,147 @@
...
   ;; Take an amount of blocks (4-byte words) to allocate, return an address
   ;; handle suitable for giving to other access methods
-  (func (export "alloc") (param $amount i32) (result i32)
+  (func $alloc (export "alloc") (param $amount i32) (result i32)
     (local $addr i32)
     (local.set $addr (global.get $heap))
     (global.set $heap (i32.add (global.get $heap) (i32.mul (local.get $amount) (i32.const 4))))
     (local.get $addr))
 
+  (func $duplicate_str(export "duplicate_str") (param $source i32) (param $dest i32)
```
jpolitz, line 15: I think these should be in a different module than `memory.wat` – that should _just_ be for the memory operations. Perhaps a new file called `string.wat`?

stdlib/strings.wat
```diff
@@ -0,0 +1,8 @@
+(module
```
jpolitz, line 1: I see – you did make this file. This could import the memory operations and export the string operations.

ir.ts
Yongxing-Chen: Hi, we update the work in the pull request of week 8 which is #60.

williammm001:
Hi Joe, our last eat modification is actually at the pr #60 rather than #40.
________________________________

stdlib/memory.wat
Yongxing-Chen: Sure.

jpolitz:
Thanks. Please feel free to close PRs that should no longer be merged.

PR #41, opened 2022-05-14
Fancy Calling Convention Design - Milestone #1

b04505009:


type-check.ts
```diff
@@ -179,165 +205,164 @@ export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<n
...
-      return {...stmt, a: NONE, obj: tObj, value: tVal};
+      return { ...stmt, a: NONE, obj: tObj, value: tVal };
   }
 }
 
-export function tcExpr(env : GlobalTypeEnv, locals : LocalTypeEnv, expr : Expr<null>) : Expr<Type> {
-  switch(expr.tag) {
-    case "literal": 
-      return {...expr, a: tcLiteral(expr.value)};
+export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<null>): Expr<Type> {
```
jpolitz: There's a lot of stuff here that's just whitespace changes. Can you revert those small changes that aren't new code? They will conflict others badly.

jpolitz (changes requested):
I think this design is pretty reasonable, and pretty straightforward. Nice work.

The complex interaction is going to be with closures (because of `Callable`). I think you should look at their PR and think about how those interactions will work. I think that is more important for a next task than any fancier kind of calling convention.

Can you please fix the whitespace issues? I'm concerned about spurious merge conflicts with future groups.

type-check.ts
b04505009: Sure. I reverted the unnecessary changes.

PR #42, opened 2022-05-14
 Generics - Milestone #1

abhishekc-sharma:
Added support for Generic Classes parameterized by globally defined Type-Variables with simple constraints. We were able to get all of our planned test cases working for Classes.

### What Should Work
- Generic Classes with any number of Type-Parameters. Type parameters can be used in fields and methods.
- Type-Variables with constraints that are simply a list of Types. The type-checker will ensure that the type actually used is a subtype of one of the types in the constraints.
- Generics can be used in field lookup, field-assignment and method calls.

### What Definitely Won't Work
- Constructing an object of a Generic Type in a deeply nested expression.
- Functions (outside any class) that use Type-Variables.
- Methods that use Type-Variables besides those that the enclosing class is parameterized by.
- Constraints on Type-Variables that are themselves generic.

As an aside - I also added a package and command to run tests and build a report of the test-coverage to get an idea of what cases my tests were actually hitting and if they were the ones I expected. Have included it in the PR since I think it is generally helpful to have.


tests/generics/e2e.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import {typeCheck} from "../helpers.test";
+
+describe('e2e tests to check generics', () => {
+    const prog0 = `
+    L = TypeVar('L')
+    R = TypeVar('R')
+
+    class Pair(Generic[L, R]):
+        left: L = __ZERO__
+        right: R = __ZERO__
```
jpolitz, line 12: Test case to try – define `swap` here that produces a `Pair[R, L]`, and related constructions.

tests/generics/monomorphizer.test.ts
```diff
@@ -0,0 +1,130 @@
+import { expect } from 'chai';
+import { Program, Type } from '../../ast';
+import { monomorphizeProgram } from '../../monomorphizer';
+
+describe('monomorphizeProgram(program) function', () => {
+    it('monomorphizes a generic class in a program with two var inits with generic type int', () => {
+        const tast : Program<Type> = { a: { tag: 'none' }, inits: [{name: 'c1', type: { tag: 'class', name: 'C', params: [ { tag: 'number' } ] },
```
jpolitz: Whoa nice. Not great to have too many of these, but some is cool.

These will be a pain to maintain when AST changes.

type-check.ts
```diff
@@ -63,19 +73,81 @@ export type TypeError = {
   message: string
 }
 
-export function equalType(t1: Type, t2: Type) {
+// combine the elements of two arrays into an array of tuples.
+// DANGER: behaves weirdly with arrays of different length.
```
jpolitz: Can we make it just throw an error if lengths are different?

type-check.ts
```diff
@@ -63,19 +73,81 @@ export type TypeError = {
...
+function zip<A, B>(l1: Array<A>, l2: Array<B>) : Array<[A, B]> {
+  return l1.map((el, i) => [el, l2[i]]); 
+}
+
+export function equalType(env: GlobalTypeEnv, t1: Type, t2: Type) {
+  // If either of t1 or t2 are type-variables, then the actual type
+  // that they are currently instantiated to needs to be fetched from
+  // the environment before comparing the types.
```
jpolitz: Yeah this seems pretty gnarly! I still wonder about monomorphizing _before_ doing this to just leverage the existing type checker. What benefit do we get from adding this whole new layer of env? Like what's a program we can handle with this approach but not by monomorphize-then-typecheck?

Or, what would be complicated in the implementation elsewhere if we didn't follow this order?

type-check.ts
```diff
@@ -63,19 +73,81 @@ export type TypeError = {
...
+export function getClassName(env: GlobalTypeEnv, t: Type): string {
+  if(t.tag === "typevar") {
+    // If t is a type-variables, then the actual type
+    // that it is currently instantiated to needs to be fetched from
+    // the environment.
+    t = env.typevars.get(t.name)[2];
+  }
+
+  if(t.tag !== "class" || t.name === UNCONSTRAINED) {
+    return undefined;
```
jpolitz: Oof I don't like using `undefined` as a meaningful value here. Very confusing. What does this mean?

type-check.ts
```diff
@@ -63,19 +73,81 @@ export type TypeError = {
...
+  }
+
+  return t.name;
 }
 
 export function isSubtype(env: GlobalTypeEnv, t1: Type, t2: Type): boolean {
-  return equalType(t1, t2) || t1.tag === "none" && t2.tag === "class" 
+  // If either of t1 or t2 are type-variables, then the actual type
+  // that they are currently instantiated to needs to be fetched from
+  // the environment before comparing the types.
```
jpolitz: I'm confused about how we know which types they are instantiated to. That phrase doesn't make sense when, say, the whole program is just a generic class.

type-check.ts
```diff
@@ -150,16 +394,163 @@ export function tcDef(env : GlobalTypeEnv, fun : FunDef<null>) : FunDef<Type> {
   return {...fun, a: NONE, body: tBody};
 }
 
+// Generic classes are type-checked by instantiating their type-parameters to every valid
```
jpolitz: Doesn't this effectively re-do all the work that you do in monomorphizing? If you have all the instantiated copies of the classes, why not just type-check those?

type-check.ts
```diff
@@ -179,6 +570,19 @@ export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<n
       } else {
         throw new TypeCheckError("Unbound id: " + stmt.name);
       }
+
+      // TODO: this is an ugly temporary hack for generic constructor
+      // calls until explicit annotations are supported.
+      // Until then constructors for generic classes are properly checked only
+      // when directly assigned to variables and will fail in unexpected ways otherwise.
```
jpolitz, line 511: Ugh. Understood. Can we throw a big error here if params.length > 0 and we _don't_ know a type to assign it to?

jpolitz (changes requested):
Generally this is great – monomorphize is really cool.

Write some more tests that swap type variables around and use them in interesting ways; that will help explore some edge cases.

See if you can make the construction case into an actual error message.

I really want to understand _why_ you're doing separate type-checking with all the environment infrastructure, and not just type-checking after expanding (if it's good enough for C++ and it's what we'll be doing anyway...). It just feels like the TC gets a huge amount of complexity and duplicated effort.

I think you're already talking to the inheritance group, but look at their PR and think about how (a) parsing and (b) subtyping will work.

type-check.ts
abhishekc-sharma: This means that we are trying to get the class name of something that is not an object. This is equivalent to the earlier explicit check for "lookup"/"method-call" where we were checking if the lhs expression has the tag class, only it gets hidden behind this function that returns the class name. 

I think I initially implemented this as a function that returned Type but this needed me to do a redundant comparison on the outside to satisfy typescript for some reason... As I write this I feel like it should be possible to clean it up in that way without requiring any redundant comparison. Let me try to do that.

type-check.ts
abhishekc-sharma: In the current implementation where we support Generic Classes only, our typechecking process for each class is -
 1. Replace all entities "wrongly" labelled as "class" but are infact type-variables with the "typevar" tag. These are all the sources of genericness within each class.
 2. Cycle through all possible combinations of assignments for the type-parameters for the class, save the assignment in the environment and run the existing `tcClass`function.

To answer your questions - 
1. We know the types they are assigned to by looking at the environment. Based on the above implementation, we should never encounter something with the type tag "typevar" outside of the context of a Generic Class being typechecked using tcClass with a particular assignment of type-variables in the environment.
2. If the whole program is a generic class, it is not a problem. The assignment in the environment is based on the constraints for the Type-Variables and are independent of the actual instantiations for the type-parameter made during the program.

type-check.ts
abhishekc-sharma: I may be misunderstanding the exact condition that you are suggesting that I check here and throw an error. I think I understand what the error/problem is - currently if a Generic Class constructor is used in some unsupported place the params.length part will never get populated, leaving some invalid type tags. Currently our best bet is to hope that it gets caught down the line in some isAssignable check (which may also not always be the case). But I feel like this is not easy to check here in single condition ? -  it would require adding checks in every place that uses and relies on the result of a tcExpr and the type is "constructor" - which I'm not sure if it is worth handling, considering that we plan to get rid of this limitation in the next milestone.

type-check.ts
abhishekc-sharma: We definitely contemplated between the two approaches -
1. Monomorphize everything before type-checking and then type-check the newly created monomorphized classes.
2. Type-check without monomorphizing and then monomorphize.

Ultimately the decision came down to a Language Design thing of how we wanted Generics to work ( which may not necessarily be the way python3 checkers do things ). Some of our thought process and understanding was influenced by https://www.tangramvision.com/blog/c-rust-generics-and-specialization. Basically our high-level conclusion was that the C++ way would to do things is 1 and the Rust way to do things is 2.

How this translates to our design choices is:
- We want a Generic Class to be type-checked and valid for all types satisfied by its constraints regardless of whether it is actually instantiated / what types it is actually instantiated with over the course of the program.

Example 1:

```python
T = TypeVar('T') # Unconstrained type-variable

class Foo(Generic[T]):
	x : T = __EMPTY__
	
	def foo(self: Foo[T], other: Foo[T]) -> T:
		return self.x + other.x
```

We want the above program to fail compilation since T is an unconstrained type variable and we don't know that the "+" operation is valid for it. Even though no objects of Foo are actually ever created. If we monomorphized before type-checking (currently we rely on variable definition's type annotations to monomorphize), there is nothing to monomorphize since we don't know any concrete types to actually instantiate T with. So Foo would disappear before type checking and we wouldn't get an error.

Example 2:

```python
T = TypeVar('T') # Unconstrained type-variable

class Foo(Generic[T]):
	x : T = __EMPTY__
	
	def foo(self: Foo[T], other: Foo[T]) -> T:
		return self.x + other.x
		
f : Foo[int] = None
```

We want this to fail for the same reasons as the previous example. Even though the case actually used in the program - 'int' is valid, we want Foo to type-check for all possible assignments/instantiations. Again if we monomorphized before type-checking, we would create only a Foo$int which would type-check and we would silently ignore the error. From a programmers point of view we were of the opinion that it is better to be strict and do this checking early rather than an error suddenly being thrown in the future when Foo[T] is instantiated for some other type that does not have the '+' operations.

- We want our Generic Classes to be "truly" monomorphized to the types they are instantiated to.

Example 3:

```python
T = TypeVar('T', bound=Superclass)

class Superclass(object):
	pass
	
class Subclass1(Superclass):
	pass
	
class Subclass2(Superclass):
	pass
	
class Foo(Generic[T]):
	x : T = __EMPTY__
	
f1: Foo[Subclass1] = None
f2: Foo[Subclass2] = None
```

This example uses the "upper bound" constraint that we hope to introduce as part of our Milestone2. In this case, we want Foo to type-check for whenever T is instantiated to any subclass of Superclass, but at the same time we want to eventual monomorphized versions of the class used to be Foo$Subclass1 and Foo$Subclass2 and not Foo$Superclass. If we monomorphized for Subclass1 and Subclass2 and then type-checked, we would potentially again fail to report an error if the code for Foo used some field/method that was present in Subclass1 and Subclass2 but not in Superclass.

Given these requirements/choices - the way we did it was simply the first that we landed on that seemed to work. Hopefully this clears things up a bit and justifies the additional complexity somewhat. Please let us know if you would still like to see any changes to reduce the amount of type-checker changes.

tests/generics/e2e.test.ts
uvashishtha: Done!

tests/generics/monomorphizer.test.ts
abhishekc-sharma: I realize I have way too many tests like this that actually use the AST for the Type-Checker..... It was the only way to test independently before the parser and monomorphizer were done. Hopefully the remaining AST changes from this point own will be trivial.

type-check.ts
abhishekc-sharma: Done

type-check.ts
abhishekc-sharma: Have still not done this since it seems like a lot of work for something we plan to fix as part of the next milestone within the next 2-3 days

abhishekc-sharma:
@jpolitz We have made the changes to remove constraints and resolved all conflicts so far. Hopefully this is in a suitable state to get merged at the earliest.

PR #43, opened 2022-05-14
for loops/ iterators- week 7

Ananya273:
Initial commit that support functionality of for loops on custom classes.

### DIFF
--- a/lower.ts
+++ b/lower.ts
@@ -2,6 +2,7 @@ import * as AST from './ast';
 import * as IR from './ir';
 import { Type } from './ast';
 import { GlobalEnv } from './compiler';
+import { BOOL } from './utils';
 
 const nameCounters : Map<string, number> = new Map();
 function generateName(base : string) : string {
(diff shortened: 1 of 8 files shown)


lower.ts
```diff
@@ -192,6 +193,38 @@ function flattenStmt(s : AST.Stmt<Type>, blocks: Array<IR.BasicBlock<Type>>, env
 
       blocks.push({  a: s.a, label: whileEndLbl, stmts: [] })
 
+      return [...cinits, ...bodyinits]
+    case "continue":
+      pushStmtsToLastBlock(blocks, { tag: "jmp", lbl:  env.labels[0]})
```
jpolitz: Will these work within while loops as well?

If so, awesome, can we add a test or two demonstrating that?

tests/forloops.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    return self
+  def next(self:Range)->int:
+    c : int = 0
+    c = self.current
+    self.current = self.current + 1
+    return c
+  def hasnext(self:Range)->bool:
+    return self.current < self.max
+ 
+def checkForLoop(n:int):
```
jpolitz, line 335: Thank you for writing a test in a function!

tests/forloops.test.ts
```diff
@@ -0,0 +1,339 @@
...
+  def hasnext(self:Range)->bool:
+    return self.current < self.max
+
+
+
+
+cls:Range = None
+i:int = 0
+cls = Range().new(1,3)
+for 1+2 in cls:
```
jpolitz: I'm genuinely surprised this is a type error and not a parse error. Is our testing util doing what we expect?

type-check.ts
```diff
@@ -212,7 +212,26 @@ export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<n
...
+    case "break":
+    case "continue":
       return {a: NONE, tag: stmt.tag};
+    case "for":
+      var tIterator = tcExpr(env, locals, stmt.iterator)
+      if(tIterator.tag!="id")
+        throw new TypeCheckError("iterator must be an id");
+      var tValObject = tcExpr(env, locals, stmt.values)
+      if (tValObject.a.tag !== "class") 
+        throw new TypeCheckError("values require an object");
```
jpolitz, line 222: Nice. This is a good place to generalize a bit when we get to lists and strings, etc.

In general I appreciate how straightforward this type-checking is. Easy for me to see how it will be extended/made more flexible in the future.

type-check.ts
```diff
@@ -212,7 +212,26 @@ export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<n
         throw new TypeCheckError("Condition Expression Must be a bool");
       return {a: NONE, tag:stmt.tag, cond: tCond, body: tBody};
     case "pass":
+    case "break":
+    case "continue":
       return {a: NONE, tag: stmt.tag};
+    case "for":
+      var tIterator = tcExpr(env, locals, stmt.iterator)
```
jpolitz: Oh interesting idea here, since the variable has to be declared earlier and have a type, you can get the type here. Note that even if you restrict it to be a string (which we probably want to do, and then maybe generalize to destructuring assignment later), you can use the environment to find its type. That would make a bit more sense to me in the long run.

jpolitz (changes requested):
Nice work. I'd like to have an answer/test for what happens with break and continue within while.

Another important test to try is a _nested_ loop, especially nesting with continue and break (since that's where those label indices will matter a lot!)

Also, it would be useful to have a test for iterating over non-ints, because it should work with the current implementation but is untested.

lower.ts
akchekur: I did not test the continue/break statements in while and nested loops. Will add those tests and necessary changes if any. Thanks for the suggestion.

tests/forloops.test.ts
akchekur: The implementation for week-7 considers iterator as an expression where 1+2 is valid. A type check on iterator to be a "id" is performed in type-check.ts due to which this testcase results in type error instead of parse error.  

The iterator in week-8 implementation is considered as string and the corresponding checks are added in the parser.

type-check.ts
akchekur: Week-8 implementation considers iterator as string and its type is computed using environment variables. Thanks for the comment.

lower.ts
akchekur: Week-8 implementation has required modifications an testcases for the break and continue statements to work in the while. loop and nested loops. Also, test cases to iterate over non-int iterators are included.  Thanks for the suggestion.

PR #44, opened 2022-05-14
Inheritance milestone1

anuj-ahuja:


ir.ts
```diff
@@ -28,7 +28,7 @@ export type Expr<A> =
   | {  a?: A, tag: "builtin1", name: string, arg: Value<A> }
   | {  a?: A, tag: "builtin2", name: string, left: Value<A>, right: Value<A>}
   | {  a?: A, tag: "call", name: string, arguments: Array<Value<A>> } 
-
+  | {  a?: A, tag: "call_indirect", index: number, arguments: Array<Value<A>> }
```
jpolitz: How can this be a constant number for `index`? Different subclasses will have their methods at different indices, right? It's fine if you focused on fields first (actually quite reasonable as the group working on closures also needs to define this, and getting the definition right will make merging easier). However, this seems like the wrong design and this has to be a `Value<A>`.

lower.ts
```diff
@@ -21,7 +21,39 @@ function generateName(base : string) : string {
...
+        } else {
+          superClassVtable[methodOffset] = `$${cls.name}$${m.name}`;
+        }
+      })
+      classIndexes.set(cls.name, [vtable.length, vtable.length +superClassVtable.length])
+      vtable = [...vtable , ...superClassVtable]
+    } else {
+      // add methods directly and increment methodIndex
+      classIndexes.set(cls.name, [vtable.length, vtable.length + cls.methods.length - 1])
+      cls.methods.filter(m => m.name !== "__init__").forEach(m => {
```
jpolitz: Why leave out `__init__`? Can't it still be called like other methods?

tests/inheritance.test.ts
```diff
@@ -0,0 +1,306 @@
...
+	b : int = 2
+
+	def sum_a_b(self: B) -> int: 
+		return self.a + self.b
+
+l : B = None 
+l = B()
+print(l.sum_a_b())
+`, [`3`]);
+assertPrint("simple-method-overriding", `
```
jpolitz, line 188: This doesn't test anything about method overriding, right? It does check that field lookup across the class hierarchy works (which is great!).

tests/inheritance.test.ts
```diff
@@ -0,0 +1,306 @@
...
+
+class B(A):
+
+    y:int = 10
+    
+a : A = None
+a = B()
+print(a.increment(1))
+`, [`2`]);
+assertPrint("method-doesnt-exist-in-super-class", `
```
jpolitz, line 126: This is the same name as another test below.

jpolitz (changes requested):
Missing an updated `design.md`, which would help understand what the checkpoint is.

From my review, it looks like this is supporting fields first, which is great, along with some of the static/environment stuff for inheriting methods. However, there is no correct generation of code for calling inherited/overridden methods (and no tests for that either).

That's fine as a checkpoint! I think your next step is just to get all the methods stuff working, and harmonizing with other groups.

I would suggest reviewing the `call_indirect` instruction in the IR, because I believe the index will have to be a value, and also looking at the closures group's PR to see how they used it (https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/45).

Please remove `.DS_Store` from this PR – we will add it to .gitignore to avoid accidentally adding it in the future.

I think the design overall is reasonable and what you've added is necessary.

tests/inheritance.test.ts
anuj-ahuja: By mistake we included the test for simple-method-call twice instead of method overriding. Have updated it to test method overriding now.

tests/inheritance.test.ts
anuj-ahuja: Removed the duplicate test case.

lower.ts
anuj-ahuja: Yes, we can do that. 

We initially thought it might be simpler to keep the constructors separate as simple function calls. But after reviewing it again now, we agree that it would be easier to handle it like methods since we won't have to filter and handle `__init__` separately every time we need to do something with class methods. 

The code is updated now.

ir.ts
anuj-ahuja: Thanks for pointing this out! 

We have updated our definition of `call_indirect` so that it works well for both closures and inheritance. It looks like this now:

`{  a?: A, tag: "call_indirect", fn: Expr<A>, name: string, methodOffset: Value<A>, arguments: Array<Value<A>> }  `.

anuj-ahuja:
We have included an updated design.md file now. We did focus on inherited/overridden methods and had most of the stuff set up for it but the definition for `call_indirect` was not correct. We have fixed it now and also included a few more test cases for method overriding.

PR #45, opened 2022-05-14
First-class functions, closures

dtjong:


ast.ts
```diff
@@ -27,19 +30,23 @@ export type Stmt<A> =
...
+export type Lambda<A> = {  a?: A, tag: "lambda", params: Array<string>, type: Callable, expr: Expr<A> };
 export type Expr<A> =
     {  a?: A, tag: "literal", value: Literal }
   | {  a?: A, tag: "id", name: string }
   | {  a?: A, tag: "binop", op: BinOp, left: Expr<A>, right: Expr<A>}
   | {  a?: A, tag: "uniop", op: UniOp, expr: Expr<A> }
   | {  a?: A, tag: "builtin1", name: string, arg: Expr<A> }
   | {  a?: A, tag: "builtin2", name: string, left: Expr<A>, right: Expr<A>}
-  | {  a?: A, tag: "call", name: string, arguments: Array<Expr<A>> } 
+  | {  a?: A, tag: "call", fn: Expr<A>, arguments: Array<Expr<A>> }
```
jpolitz, line 56: Oof merge conflict alert. I wonder if there's a way to split this into a string version and an expression version? Definitely at least the folks working on calling conventions will conflict with this, probably others.

compiler.ts
```diff
@@ -172,6 +179,16 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
       return valStmts;
 
+    case "call_indirect":
+      if (expr.fn.a.tag !== "callable") {
+        throw new Error("cursed");
+      }
+      
+      var valStmts = codeGenValue(expr.fn, env);
+      var fnStmts = expr.arguments.map((arg) => codeGenValue(arg, env)).flat();
+      // +1 for self
```
jpolitz: I like the idea of `call_indirect` on its own to the IR.

I'm not sure I like this design for it, though. Since you're just doing a `load` of `0` offset, why not represent this in lower? You could also capture the extra +1 argument there.

The compilation of `call_indirect` would then not even be responsible for `load` or making assumptions about where the offset should be, and would be more re-usable. The type could be computed similarly _or_ be a part of the `call_indirect` IR form itself.

compiler.ts
```diff
@@ -286,7 +304,7 @@ function codeGenDef(def : FunDef<Type>, env : GlobalEnv) : Array<string> {
 
 function codeGenClass(cls : Class<Type>, env : GlobalEnv) : Array<string> {
   const methods = [...cls.methods];
-  methods.forEach(method => method.name = `${cls.name}$${method.name}`);
+  methods.forEach(method => method.name = createMethodName(cls.name, method.name));
```
jpolitz, line 354: Nice to put this in a helper, thanks.

designs/closures-design.md
```diff
@@ -0,0 +1,362 @@
...
+
+## What we implemented
+
+We implemented what we committed to with the exception of free and nonlocal variables. We 
+mostly stuck to our original plan.
+
+If expressions are fully implemented, however once inheritance is implemented
+more work may need to be done.
+
+First class functions are implemented except for methods.
```
jpolitz, line 11: I don't understand what "except for methods" means here.

ir.ts
```diff
@@ -28,7 +28,7 @@ export type Expr<A> =
   | {  a?: A, tag: "builtin1", name: string, arg: Value<A> }
   | {  a?: A, tag: "builtin2", name: string, left: Value<A>, right: Value<A>}
   | {  a?: A, tag: "call", name: string, arguments: Array<Value<A>> } 
-
+  | {  a?: A, tag: "call_indirect", fn: Value<A>, arguments: Array<Value<A>> }
```
jpolitz: Connected to my other comment in `compiler.ts`, I like this definition of `call_indirect`! I just think in `lower.ts` you should add any extra args and do the appropriate load ahead of time.

tests/parser.test.ts
```diff
@@ -1,44 +0,0 @@
-// import * as mocha from 'mocha';
```
jpolitz: Why delete these?

jpolitz (changes requested):
Broadly, this is great. I'm nervous about the merge with inheritance, but we can manage that.

The main thing I want to hear back about (and potentially see changed) is a simpler compilation of `call_indirect`. Why can't it just use the value of the first argument as the table offset, and compute the type directly from the number of arguments? Leave `lower` to do a little more work.

Great job testing, and I get the overall design.

designs/closures-design.md
pashabou: We don't currently support returning or storing references to methods (specifically, class methods), just top-level or nested functions. We'd have to think about the calling syntax and representation of bound/unbound methods, e.g. `f = Cls.func` vs. `f = Cls().func`.

tests/parser.test.ts
dtjong: Accident, originally started testing here but then moved to new file to prevent merge conflict, forgot to checkout again.

ast.ts
dtjong: We've been talking to the calling conventions team and they say they can work with this version of the AST. We'll keep in touch with them and see if anything changes.

We've also looked through the other PRs and we don't think this change will interfere with any of the other groups severely.

PR #46, opened 2022-05-14
Builtins Milestone 1

LER0ever:
# DEPRECATED! DO NOT MERGE, SUPERCEDED BY #101 

Builtins Proposal: #10


compiler.ts
```diff
@@ -1,25 +1,31 @@
 import { Program, Stmt, Expr, Value, Class, VarInit, FunDef } from "./ir"
 import { BinOp, Type, UniOp } from "./ast"
-import { BOOL, NONE, NUM } from "./utils";
+import { BOOL, NONE, ELLIPSIS, NUM, FLOAT } from "./utils";
+import { errorMonitor } from "events";
```
jpolitz, line 4: This import is unused.

compiler.ts
```diff
@@ -1,25 +1,31 @@
...
   labels: Array<string>;
   offset: number;
 }
 
 export const emptyEnv : GlobalEnv = { 
   globals: new Map(), 
+  globalfloats: new Map(),
   classes: new Map(),
   locals: new Set(),
+  localfloats: new Set(),
```
jpolitz: Whoa lots of floats going on here.

lower.ts
```diff
@@ -288,7 +304,7 @@ function flattenExprToExpr(e : AST.Expr<Type>, env : GlobalEnv) : [Array<IR.VarI
     case "id":
       return [[], [], {tag: "value", value: { ...e }} ];
     case "literal":
-      return [[], [], {tag: "value", value: literalToVal(e.value) } ];
+      return [[], [], {tag: "value", value: {a:e.a, ...literalToVal(e.value)} }];
```
jpolitz, line 317: Thanks for the fix!

parser.ts
```diff
@@ -1,15 +1,22 @@
...
 import { Program, Expr, Stmt, UniOp, BinOp, Parameter, Type, FunDef, VarInit, Class, Literal } from "./ast";
-import { NUM, BOOL, NONE, CLASS } from "./utils";
+import { NUM, FLOAT, BOOL, NONE, ELLIPSIS, CLASS } from "./utils";
 import { stringifyTree } from "./treeprinter";
 
 export function traverseLiteral(c : TreeCursor, s : string) : Literal {
   switch(c.type.name) {
     case "Number":
+      const tonum = Number(s.substring(c.from, c.to));
+      if (tonum !== Math.floor(tonum)){
```
jpolitz: What about floats like `Infinity` or `1e100` where this is true, but they are not representable as ints?

I think this float stuff needs to be taken out in its current form and redesigned somewhat; this isn't a safe way to tell if something is a float, more metadata is needed.

parser.ts
```diff
@@ -57,13 +69,19 @@ export function traverseExpr(c : TreeCursor, s : string) : Expr<null> {
       } else if (callExpr.tag === "id") {
         const callName = callExpr.name;
         var expr : Expr<null>;
-        if (callName === "print" || callName === "abs") {
+        if (callName === "print") {
```
jpolitz, line 93: I think as the builtins team, your job should be to see if there are ways to _remove_ these hard-coded lists of `||` conditions, and make a more general solution possible. What are some options?

tests/pa3-hidden.test.ts
```diff
@@ -1,325 +1,325 @@
-import { assertPrint, assertTCFail, assertTC, assertFail } from "./asserts.test";
```
jpolitz: Why the large diff on this file?

type-check.ts
```diff
@@ -169,6 +172,9 @@ export function tcBlock(env : GlobalTypeEnv, locals : LocalTypeEnv, stmts : Arra
 
 export function tcStmt(env : GlobalTypeEnv, locals : LocalTypeEnv, stmt : Stmt<null>) : Stmt<Type> {
   switch(stmt.tag) {
+    case "import":
```
jpolitz, line 175: Where are imports actually handled?

jpolitz (changes requested):
There's a lot going on here!

How are the `.py` files and `.wat` files connected to the programs we run?

It seems like there's a bunch of code to parse imports, but I don't see where they are actually used/processed in programs?

And the stuff about float seems like an entire design that has some issues.

I'm confused that there aren't any tests that use `import`, and about the commented-out tests.

I would refocus on:

- A simple set of tests that uses imports
- Getting rid of the float stuff
- A clear way to turn any libraries you want to implement in .wat/.py into importable things
- A solution for all the hardcoded names of builtins in compile/lower

LER0ever:
Re: https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/46#pullrequestreview-977366265 @jpolitz 
Hi Prof. Politz, 


tldr: For `import` statement, milestone 1 implements a basic parser support, and all the built-in functions are available regardless of whether they are imported or not. Milestone 2 plans to add a basic import control and a more generic way of adding built-in functions, so that we don't have to hardcode built-in checks and import would actually behave normally.

---------

we plan to implement a basic but generic import support to handle all the built-in functions we add. Currently this milestone has built-in function available in the namespace. And the import statement only have compiler frontend support, i.e. import statements are correctly parsed, but turned into `<nop>` in IR lowering, essentially having no effects on the code execution.

```python
# This works
from math import abs
print(abs(5))
# This also works for now, without import
print(gcd(6, 3))
```

In the upcoming weeks, we plan to use the import statement as a switch for built-in functions. None of them are available to use unless explicitly imported.

In the lowering and codegen process, the corresponding typescript-wasm bridging functions are only added when this function is imported.

We plan to maintain all the built-in functions in a map, with their name, type signatures and the typescript implementations, and dynamically dispatch some of them to codegen according to import statements, allowing us to implement dozens of functions without hardcode the `callName` and number of arguments in parser and runner.

> How are the .py files and .wat files connected to the programs we run?

We are still exploring both options mentioned in #10 , and could really use some expertise.

parser.ts
LER0ever: We plan to move towards a more generic built-in function implementation that does not require these hardcoding. Ref: https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/46#issuecomment-1135492651

type-check.ts
LER0ever: Import statements currently only have parser support (just to make sure we pass the tests listed in the first week), and have no effects on lowering and codegen. We are proposing a new simple import for the next milestone, but would inevitably hit the same problem in the original proposal: #10 . We can probably further improve our milestone 2 after the class tomorrow.

tests/pa3-hidden.test.ts
LER0ever: We commented out the file for debugging purposes, and have now reverted the change.

PR #47, opened 2022-05-15
Error reporting milestone 1

rlisahuang:


.gitignore
```diff
@@ -5,4 +5,6 @@ debugger
 cli
 *.sw*
 mochawesome-report
-*lock*
\ No newline at end of file
+*lock*
+*.DS_Store
+package-lock.json
```
jpolitz, line 9: Yeah probably a good idea.

jpolitz (commented):
This looks pretty good. I'd appreciate some more testing, especially a few tests that check source locations, because there's a lot of infrastructure here that needs to be made to work.

However, I generally think the `Annotation` stuff and way errors are created is good design so I'm inclined to merge this early so other groups can work with this new structure.

The main feedback I have is to get some more tests in place to make sure we don't break errors going forward.

An interesting design question is whether we can avoid calling into TS/JS so often when doing the asserts, and measuring if that has a really bad performance impact.

jpolitz:
Can you all look into why these are failing?

https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/runs/6494285931?check_suite_focus=true

madun1999:
> Can you all look into why these are failing?
> 
> https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/runs/6494285931?check_suite_focus=true

Current tests are fixed. Sorry about that.

.gitignore
Michaelmvv: While in my personal experience, ignoring the `package-lock.json` file has never caused a problem, ignoring it in git is not recommended in the official documentation. I would consider the following before merging:

[npm's documentation on package-lock.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json) states that one of the main reasons to commit it is that it's used to
> Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.

[Node's documentation](https://nodejs.dev/learn/the-package-lock-json-file) states that "The package-lock.json file needs to be committed to your Git repository"

I think it is fine to add it to the git ignore, but we have to remember that if/when people `npm install` after any changes to `package.json`, they could install different versions of dependencies than everyone else. There is a very slim possibility this could cause issues.

madun1999:
> This looks pretty good. I'd appreciate some more testing, especially a few tests that check source locations, because there's a lot of infrastructure here that needs to be made to work.

We know we only covered one of the Typecheck error cases. We will create more tests about these, as well as errors that occur deep in the tree.
Other than that, are there any specific test examples that you are thinking about that "check source locations"? What behavior might be broken with the new infrastructure? 

> An interesting design question is whether we can avoid calling into TS/JS so often when doing the asserts, and measuring if that has a really bad performance impact.

Certainly. The solution would be to do `assert_not_zero` inside WASM and only call into TS/JS when the assertion fails.

.gitignore
rlisahuang: Sounds reasonable. I added `package-lock.json` back to git.

jpolitz: Agree – it's good to have in .gitignore (because it stops accidental huge diffs), but I'd like it to be present in git. I was just going to regenerate it once after merging to get a good copy, but adding it back is great, too, thanks!

rlisahuang:
The last commit to the `2022` branch of our fork includes API updates to the `TypeCheckError` class in `errors.ts` on the request from the front-end team. The changes we provide would allow them to access source code and location information in `webstart.ts`, as they'd like to use such information for code highlighting etc. in the editor.

.gitignore
madun1999: We were using different versions of `npm` and they generates different `package-lock.json`. We'll ignore it until a good copy of it is generated.

Michaelmvv: >  it's good to have in .gitignore (because it stops accidental huge diffs), but I'd like it to be present in git. I was just going to regenerate it once after merging to get a good copy

Ah, I see. I did not consider this. I think you are right that keeping it in the .gitignore and regenerating it after merging is a good way of keeping `package-lock.json` up to date for everybody. 

Sorry for creating more work for you @rlisahuang and @madun1999

PR #49, opened 2022-05-19
Destructuring Assignment Milestone 1

zhuo-zhi:
This is the new pull request for destructuring assignment milestone 1.


parser.ts
```diff
@@ -218,28 +233,62 @@ export function traverseStmt(c : TreeCursor, s : string) : Stmt<null> {
...
+      //   }
+      // } else if (target.tag === "id") {
+      //   return {
+      //     tag: "assign",
+      //     name: target.name,
+      //     value: value
+      //   }  
+      // } else {
+      //   throw new Error("Unknown target while parsing assignment");
+      // }
```
KamranAlipour: Stray code, please clean up.

pancerZH: > Stray code, please clean up.

done!

tests/destructure.test.ts
```diff
@@ -0,0 +1,228 @@
...
+    class cl(Object):
+        f1: int = 0
+    
+    c = cl()
+    c.f1, _, a = range(1, 4)
+    print(c.f1)
+    print(a)
+    `, ['1', '3']);
+    
+});
```
KamranAlipour, line 255: Great tests, should we have some that check length mismatch in assign? something like:  `a,b = 1,2,3` or `a,b,c = 1`
Also going above literals, like `a,b,c = 1 , f(12), g(4)+5`

parser.ts
```diff
@@ -185,6 +185,21 @@ export function traverseExpr(c : TreeCursor, s : string) : Expr<null> {
         tag: "id",
         name: "self"
       };
+    case "ArrayExpression":
```
KamranAlipour: make sure you always have some instruction within the cases even if you don't know the implementation at the moment, for example raise an error or warning message for now, until later fill the gaps. it would be nice to leave tiny comments of a tiny example of what you are parsing in each case beneath it.

parser.ts
```diff
@@ -185,6 +185,21 @@ export function traverseExpr(c : TreeCursor, s : string) : Expr<null> {
...
+    case "TupleExpression":
+      let elements: Expr<null>[] = [];
+      c.firstChild();
+      c.nextSibling();
+      while (s.substring(c.from, c.to).trim() !== "]") {
```
KamranAlipour, line 192: is this case for tuples? like `(1,2,3)` or without brackets like `1,2,3` ? I'm a bit confused as why we are looking for `]` then?

ast.ts
```diff
@@ -40,6 +51,9 @@ export type Expr<A> =
   | {  a?: A, tag: "index", obj: Expr<A>, index: Expr<A> }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  // array-expr should be plain format like 1, 2, 3 without brackets
+  // TODO: should we make use of AST nodes from list and tuple groups?
```
KamranAlipour, line 55: eventually, I think you'll be dealing with `(1,2,3)` , `[1,2,3]`, and `1,2,3`. Am I right?

tests/destructure.test.ts
pancerZH: Thank you for comments! We do have one test for `a, b = 2`, and we can add more tests like `a, b = 2, 3, 4` and the case you mentioned.

ast.ts
pancerZH: Yes!

ast.ts
```diff
@@ -28,6 +28,17 @@ export type Stmt<A> =
   | {  a?: A, tag: "if", cond: Expr<A>, thn: Array<Stmt<A>>, els: Array<Stmt<A>> }
   | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
 
+// isSimple should be true when destruct has no comma(,)
+// e.g. a, = 1, -> isSimple = false
+// e.g. a = 1   -> isSimple = true
```
KamranAlipour, line 33: I have an idea, do you think its possible to trim down things like this in parsing and make everything simplified so you won't need `isSimple` at all? I'm a bit worried that's gonna create an extra unnecessary flag for you which might not be really beneficial... For example in the example in your comment, imagine `a,` getting automatically simplified to `a` in parsing and also `1,`  to `1`.

KamranAlipour (commented):
Great progress! I had a few questions and suggestions about your design and tests for which I left comments throughout the code.

tests/destructure.test.ts
pancerZH: > Great tests, should we have some that check length mismatch in assign? something like: `a,b = 1,2,3` or `a,b,c = 1` Also going above literals, like `a,b,c = 1 , f(12), g(4)+5`

done!

liusitan:
Hi there, I checked your design doc and test code and I am confused about what the element with * refer to, may you give me an example?

pancerZH:
> Hi there, I checked your design doc and test code and I am confused about what the element with * refer to, may you give me an example?

In Python, a, *b, c = 1, 2, 3, 4 would be interpreted to a = 1, b = [2, 3] and c = 4

parser.ts
csx2hen: In fact, I intentionally left this case empty in order to fall-through to the next case, TupleExpression, which has the similar logic. I am sorry that it made you confused. I have separated them clearly and added some comments.

parser.ts
csx2hen: That's a bug. The reason is that I wanted to reuse the logic between TupleExpression and ArrayExpression. I have separated this logic to make it clear and fixed the bug.

ast.ts
pancerZH: After careful consideration, I found it perhaps not suitable to remove `isSimple` field. See the below snippet:
```Python
>>> a, = [1]
>>> a
1
>>> a = [1]
>>> a
[1]
```
Without `isSimple` field, we could not distinguish the two scenarios.

pancerZH:
@jpolitz Hi professor, would you mind finishing merge of our pr ASAP? The generic team has talked with us and they would make use of some of our implementations, while we are going to make use some of contents of list team, which has not been merged yet. In this way, if we merge the main branch and pr from list team, and then make the generic team merge our pr, this could be a total of mess. So may we firstly keep the same pace with main branch?

pancerZH:
Please **DO NOT** merge this pr. See m2 instead.

jpolitz:
Closing per comment above

liusitan:
Got it thank you!

Best,
Sitan Liu


________________________________

PR #50, opened 2022-05-19
Frontend update for week8

liusitan:
this week we add syntax highlight, line number to our code pad. we provide an example code dropdown list, a convenient save/upload/clear button, and also the autocompletion in reply in the frontend, you may want to add an example code for your features. We currently print out the object plainly and look forward to including your data structures(string/set/list). Excited to listen to and take your comments/feedback/suggestions!


const.ts
```diff
@@ -0,0 +1,106 @@
...
+    "def",
+    "for",
+    "lambda",
+    "try",
+    "int",
+    "bool",
+    "str",
+];
+
+export const default_functions = [
```
KamranAlipour, line 38: Where does this list come from? do we need to check in with the built-in teams and somehow load this instead of hardcoding? just wondering

style.scss
```diff
@@ -0,0 +1,230 @@
...
+      display: block;
+    }
+  }
+}
+.save_load {
+  width: 100%;
+  height: 100%;
+  float: left;
+}
+// #ls {
```
KamranAlipour: if you don't need the stray code please make sure to clean up

autocomplete.ts
```diff
@@ -0,0 +1,127 @@
...
+            base = obj.string;
+        } else if (obj.type == "variable-3") {
+            base = ":" + obj.string;
+        } else if (obj.type == "property") {
+            base = obj.string;
+        } else if (obj.type == "builtin") {
+            base = obj.string;
+        } else if (obj.type == "keyword") {
+            base = obj.string;
+        }
```
KamranAlipour, line 64: seems like this many `else if`s might not be necessary here. ;)

autocomplete.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    if (context) {
+        // If this is a property, see if it belongs to some object we can
+        // find in the current environment.
+        var obj = context.pop(),
+            base;
+
+        if (obj.type == "variable") {
+            base = obj.string;
+        } else if (obj.type == "variable-3") {
```
KamranAlipour: whoa, I have no idea what "variable-3" is.

design/frontend-design.md
```diff
@@ -0,0 +1,178 @@
...
+    // in REPL:
+    >>> c
+
+    output:
+        {
+            address: 0x1234
+            a: 1
+            e: none
+        }
+```
```
KamranAlipour, line 103: good tests, but in general, I think you should test more than one operation in REPL, for example, do a couple of things in REPL and change some field values and then get an output.

KamranAlipour (commented):
Nice work! I really like the autocomplete functionalities, excited to see how they show up on the interface. Aside from the nice features, I wonder once we have lists and strings, how your design would handle super long arrays and strings? specially in your indented view of objects, if we have multiple inner layers of objects like:
` output:
        {
            address: 100
            a: True
            e: {
                address: 108
                a: 1
                b:{
                     address: 100
                     e: {
                           address: 108
                           a:{
                               address: 116
                               a:{
                                   address: 120 
                                   a:{
                                       ....[lots of inner object layers]{
                                                                                      d: some literal with long representation
                                                                                      }
                                  }
                              }
                         }
            }
        }`

const.ts
Gandalfthegreat: this is the default built-in function from the python official website.

style.scss
Gandalfthegreat: Got it! thanks

autocomplete.ts
Gandalfthegreat: Thanks, I would update

autocomplete.ts
Gandalfthegreat: Thanks, I would update

design/frontend-design.md
liusitan: tested

liusitan:
> 
We updated our design, now an object is presented as a foldable element, with a dynamically generated HTML element.

PR #60, opened 2022-05-24
Strings -- Final Submission

williammm001:


type-check.ts
```diff
@@ -274,8 +282,21 @@ export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<Anno
       const tBin = { ...expr, left: tLeft, right: tRight };
       switch (expr.op) {
         case BinOp.Plus:
+          if(equalType(tLeft.a.type, STRING) && equalType(tRight.a.type, STRING)){
+              // string concatenation is not a simple BinOp
+              return{a:{ ...expr.a, type:STRING},tag:"str-concat",left:tLeft,right:tRight};
```
jpolitz: I'm still confused about why we need new ast forms for this. Why can't this turn into a `call` to a builtin, or just record the types and handle this in lowering?

williammm001: Fixed. Now there is no extra codes in ast and ir.

PR #67, opened 2022-05-24
List - Project Milestone 2

Hackerry:
List Milestone 2 PR


lower.ts
```diff
@@ -291,6 +344,101 @@ function flattenExprToExpr(e : AST.Expr<Annotation>, env : GlobalEnv) : [Array<I
...
+        a: {...idxExpr.a, type: {tag: "number"}},
+        tag: "binop",
+        op: AST.BinOp.Plus,
+        left: e.index,
+        right: {a: {...idxExpr.a, type: {tag: "number"}}, tag: "literal", value: {tag: "num", value: 1}}
+      };
+      var [idxAddInit, idxAddStmts, idxAddExpr] = flattenExprToVal(indexAdd, env);
+
+      return [
+        [...objInit, ...idxInit, ...idxAddInit,
```
jpolitz, line 427: I'm concerned about this, because it suggests that if we had something like `lst[f()]`, then we'd evaluate `f()` twice. Can you check on that?

jpolitz:
See https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/tree/list-merge-attempt

@zhuo-zhi there are some pretty significant conflicts with the use of list construction expressions with destructuring assignments. I didn't review destructuring in as much detail, but there's definitely an issue here.

The more obvious issue is with bignums and indices, which we knew would happen @taipeiyuka – what's the easiest way for us to generate a statement/expression that converts a bignum to a regular i32 in lower?

Could we maybe make the compiler automatically do that for uses of bignums in load/store, since we don't really ever want those to be bignums?

pancerZH:
> list construction expressions

Looks like there is a conflict for parsing **ArrayExpression**. We processed the same tag:
```typescript
    case "ArrayExpression":
      const elements = traverseArray(c, s, env);
      // if there are multiple brackets like [1,a][x], treat first section as array construction and second as index
      return {
        tag: "construct-list",
        items: elements,
      };

    case "ArrayExpression": // a, b, c = [1, 2, 3]
      let arrayElements: Expr<Annotation>[] = [];
      c.firstChild();
      c.nextSibling();
      while (s.substring(c.from, c.to).trim() !== "]") {
        arrayElements.push(traverseExpr(c, s, env));
        c.nextSibling();
        c.nextSibling();
      }
      c.parent();
      return {
        tag: "array-expr",
        elements: arrayElements,
      };
```
As well as the definition in AST:
```typescript
| {  a?: A; tag: "array-expr", elements: Array<Expr<A>> }
| {  a?: A, tag: "construct-list", items: Array<Expr<A>> }
```
Basically the same thing.
And there could also be overlapping efforts in type checker, lower and compiler.

taipeiyuka:
Bignums: 
In this pull request https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/pull/85, we are using a helper function called `$bignum_to_i32` to convert the index and the post-binop bignum into `wasmint`'s in `lower.ts`. There are still some weird behaviors, but hopefully it gives the taste of how we are dealing with the issue.

PR #80, opened 2022-05-24
Bignums milestone1 merge conflict patch

taipeiyuka:
We created conflict in `webstart.ts` which was unable to detect by `npm test`. Now the code is fixed.


taipeiyuka:
The original IR does not return negative values in `num` type. It is converted by unary operators. After the optimization, they could have passed negative values as values into `codeGenValue`. This conflict is now fixed.

taipeiyuka:
We fixed the bug that was shown in class where it cannot handle some number larger than a given length. It is corrected by letting `parser` return a `BigInt` instead of `Number` for type `"num"`.

PR #101, opened 2022-06-02
Builtins Milestone 2: Final PR, No Conflict!

LER0ever:
## Built-in Libraries, Import Statements, Float, Print
This PR supersedes all of our previously unmerged PRs: #10 , #46 , #77 . We've consolidated all our previous changes (from both milestone 1 and 2) to this single pull request, which **currently has no merge conflict with `2022` branch and passes all the unit tests**!

_The previous 3 PRs are therefore now deprecated, please merge this one instead. It's ready to land in 2022!_

## Float Type
```python
pi: float = 3.1415926
def circle_area(r: float) -> float:
    return pi * r * r
print(circle_area(2.0))
```

## Import Statement
We implemented a simple import statement to control which builtin functions are enabled.

```python
from module import function
```

## Builtin Library
See all the builtin function database here at [builtins.ts](https://github.com/ucsd-cse231-s22/chocopy-wasm-compiler-A/blob/8ed70c9d947350ab6ac26d04b7f21f3b667126ca/builtins.ts)
#### Builtin Functions
These functions are available globally, works with bigint, and do not require import to use
```python
print(pow(abs(-2), 3))
print(max(1, min(2, 3)))
```

#### Math Library
These functions requires import to use, and works with bigint. For example:

```python
from math import perm
print(perm(100, 20))
```

#### Float Math Library
We separate the math lib into bigint and float, since we cannot access generics and these two types have different loading mechanisms.

```python
from float_math import log
from float_math import sqrt
print(log(1024.0, 2.0))
print(sqrt(1024.0))
```

#### Random Library
The following code for example generates a random number between factorial 20! and 40!, both are really huge.
```python
from random import randint
from math import factorial
print(randint(factorial(20), factorial(40)))
```

#### Other library
Adding new modules and functions are fairly easy, just implement the function in TypeScript and bridge them to ChocoPy by adding them to the builtin function database.

We originally planned to implement more interesting builtin libraries such as `path`, `re`, `time`, etc. But those heavy rely on the implementation from the Strings & IO group, neither of which are merged unfortunately.


LER0ever:
## Update since we filed the PR

### NOTICE on reversion of commit: 6a876c3 "a few unit tests + functionality"

6a876c3 (and thus a039ed7) was reverted because most JS math functions registered in that commit is not working and not unit-tested.

**39** Javascript math functions were registered to builtin functions, but only **3** of them have unit tests, **20**-ish of them use Float arguments, which we currently do NOT support. Several bigint functions have no usecases, such as `fabs_int`(same as `abs`), `floor_int` and `ceil_int`.

UPDATE: **Part of that commit was added back in 89c07b7** , which includes  `math.perm` and `math.comb`, with proper tests written both for small numbers and bigints.

### Support Float Type in Function parameters and returns
The fix from 7d023f2 and  8551df3 enabled us to use floating point in functions. Therefore we added back a `float_math` library, which provided builtin math functions for float, such as `sin` and `sqrt`, in commit 8ed70c9.

